Issue #231, opened 2023-02-14
Firebase iOS SDK version 10+ fails to compile


AttilaTheFun:
I'm getting this issue when trying to use firebase analytics:
```
logan@Logans-MBP Snag % bazel build //Services/Sources:AnalyticsServiceImplementation
INFO: Repository swiftpkg_firebase_ios_sdk instantiated at:
  /Users/logan/Developer/Swift/Apps/Snag/WORKSPACE:53:19: in <toplevel>
  /Users/logan/Developer/Swift/Apps/Snag/swift_deps.bzl:62:18: in swift_dependencies
Repository rule swift_package defined at:
  /private/var/tmp/_bazel_logan/9c5dc880d421e060978209038345ac60/external/cgrindel_swift_bazel/swiftpkg/internal/swift_package.bzl:167:32: in <toplevel>
INFO: repository @swiftpkg_firebase_ios_sdk' used the following cache hits instead of downloading the corresponding file.
 * Hash 'd5098cd2d731104dddb11c39b2cbbd3aab11a604c528706a16ff0114abd8d53a' for https://dl.google.com/firebase/ios/swiftpm/10.4.0/FirebaseAnalytics.zip
If the definition of 'repository @swiftpkg_firebase_ios_sdk' was updated, verify that the hashes were also updated.
ERROR: An error occurred during the fetch of repository 'swiftpkg_firebase_ios_sdk':
   Traceback (most recent call last):
        File "/private/var/tmp/_bazel_logan/9c5dc880d421e060978209038345ac60/external/cgrindel_swift_bazel/swiftpkg/internal/swift_package.bzl", line 65, column 31, in _swift_package_impl
                repo_rules.gen_build_files(repository_ctx, pkg_info)
        File "/private/var/tmp/_bazel_logan/9c5dc880d421e060978209038345ac60/external/cgrindel_swift_bazel/swiftpkg/internal/repo_rules.bzl", line 78, column 55, in _gen_build_files
                bld_file = swiftpkg_build_files.new_for_target(repository_ctx, pkg_ctx, target)
        File "/private/var/tmp/_bazel_logan/9c5dc880d421e060978209038345ac60/external/cgrindel_swift_bazel/swiftpkg/internal/swiftpkg_build_files.bzl", line 21, column 40, in _new_for_target
                return _clang_target_build_file(repository_ctx, pkg_ctx, target)
[...]

Issue #208, opened 2023-02-09
The `@import` of Objective-C modules in Objective-C does not work.


cgrindel:
I thought that I had implemented a workaround in #169. However, I can reproduce the issue by building the `examples/firebase_example/appdistribution` example code.

```
ERROR: /private/var/tmp/_bazel_chuck/61ce8061fa830ba4fbf43d9ec3d51831/external/swiftpkg_firebase_ios_sdk/BUILD.bazel:7241:13: Compiling FirebaseAuth/Sources/Backend/FIRAuthBackend.m failed: (Exit 1): wrapped_clang failed: error executing command (from target @swiftpkg_firebase_ios_sdk//:FirebaseAuth_Sources_FirebaseAuth_Objc)
  (cd /private/var/tmp/_bazel_chuck/61ce8061fa830ba4fbf43d9ec3d51831/sandbox/darwin-sandbox/679/execroot/firebase_example && \
  exec env - \
    APPLE_SDK_PLATFORM=iPhoneSimulator \
    APPLE_SDK_VERSION_OVERRIDE=16.2 \
    PATH=/bin:/usr/bin:/usr/local/bin \
    XCODE_VERSION_OVERRIDE=14.2.0.14C18 \
    ZERO_AR_DATE=1 \
[...]

PR #233, opened 2023-02-16
feat: introduce `generate_modulemap` to support `@import` in Objective-C modules

cgrindel:
Related to #231.


cgrindel:
Related to #208.

Issue #285, opened 2023-03-16
`tidy_check` fails randomly because of modified `swift_deps_index.json` 


tingyao-ui:
## Symptom
I was trying `bazel run tidy_check` commands and found an issue here.
In the example `pkg_manifest_minimal`, the generated file `swift_deps_index.json` will be randomly modified at values of `direct_dep_identities`.
e.g.
```json
{
  "direct_dep_identities": [
    "my_local_package", // <---- it's not in identical order after update-repo
    "swift-argument-parser",
    "swift-log",
    "swiftformat"
  ],
  "modules": ["..."],
  "products": ["..."]
```
<img width="638" alt="CleanShot 2023-03-16 at 14 19 53@2x" src="https://user-images.githubusercontent.com/101082831/225531875-b41b6a3f-2823-4fae-be60-c07cd4a71386.png">

## Expected Result
The `tidy_check` should not accidentally updates `swift_deps_index.json`, and `swift_update_pkgs` updates the `swift_deps_index.json` content only when the user modifies `Package.swift`.

## Root cause
`DependencyIndex.directDepIdentities` is a type of `mapset` that is not an ordered collection here.
Will it be better if we check duplicates and sort them inside `func (di DependencyIndex) AddDirectDependency`?

For instance:
```go
type DependencyIndex struct {
	moduleIndex         ModuleIndex
	productIndex        ProductIndex
	modToPrdsIndex      ModuleToProductsIndex
	prdToModsIndex      ProductMembershipsIndex
	directDepIdentities []string // Change from mapset.Set[string]
}

func (di *DependencyIndex) AddDirectDependency(identities ...string) {
	di.directDepIdentities = mapset.NewSet(append(di.directDepIdentities, identities...)...).ToSlice()
	sort.Strings(di.directDepIdentities)
}
```

Here are my total code changes for your reference
https://github.com/cgrindel/swift_bazel/compare/main...erichsu:swift_bazel:fix/eric-sort-direct-deps

tingyao-ui:
Hi @cgrindel, no problem, I'll try to create a pull request for further discussion of implementation details.

tingyao-ui:
@cgrindel I found there is a commit that already resolves this issue, I'll close the draft PR and wait for the next release.
https://github.com/cgrindel/swift_bazel/blob/12d0e9d4d5c7412cde1419cc765d63c0e3a10726/gazelle/internal/swift/dependency_index.go#L73-L80

[...]

PR #291, opened 2023-03-20
feat: sort `DependencyIndex.directDepIdentities` for swift_deps_index.json

tingyao-ui:
## Summary

PR for issue #285. 
The JSON output of `DependencyIndex.directDepIdentities` will be in random order because of the type `mapset`.
I suggest we change it to an array. 
Sort the order and remove duplicate items via `mapset` during updating.

## Implementation Details

- Change `DependencyIndex.directDepIdentities` type from `mapset.Set[string]` to `[]string`
- Sort `DependencyIndex.directDepIdentities`
- Eliminate explicit type warns


tingyao-ui:
No need to this change since `main` branch already resolve the issue #285

cgrindel:
[Release 0.3.3](https://github.com/cgrindel/swift_bazel/releases/tag/v0.3.3) has just been published. It should have the fix.

PR #364, opened 2023-04-27
Fix local_swift_package relative path resolution

### LINKED ISSUE #363: Local packages do not work with non-root based setups
If you configure the project like the [Vapor example](https://github.com/cgrindel/rules_swift_package_manager/tree/main/examples/vapor_example) and try to depend on a local package from the `Package.swift` like:

```swift
.package(path: "../Path/To/MyPackage"),
```

Then the `swift_dependencies` declaration for `local_swift_package` gets generated like so:

```python
    local_swift_package(
        ...
        path = "../Path/To/MyPackage",
    )
```

When this rule is built it [uses a relative path from the `WORKSPACE`](https://github.com/cgrindel/rules_swift_package_manager/blob/main/swiftpkg/internal/local_swift_package.bzl#L36-L41) since thats what repository contexts allow. This ends up with a failure during build.

The `path` attribute should be consumed relative to the `Package.swift` but generated relative to `WORKSPACE` file.

luispadron:
This fixes the relative path resolution for local_swift_package when the Package.swift is not in the root of the repository.

Closes #363

### DIFF
--- a/gazelle/update_repos.go
+++ b/gazelle/update_repos.go
@@ -150,7 +150,7 @@ func importReposFromPackageManifest(args language.ImportReposArgs) language.Impo
 	for _, bzlRepo := range bzlReposByIdentity {
 		repoUsage[bzlRepo.Name] = true
 		var err error
-		result.Gen[idx], err = swift.RepoRuleFromBazelRepo(bzlRepo, sc.DependencyIndexRel, pkgDir)
+		result.Gen[idx], err = swift.RepoRuleFromBazelRepo(bzlRepo, sc.DependencyIndexRel, pkgDir, c.RepoRoot)
 		if err != nil {
 			result.Error = err
 			return result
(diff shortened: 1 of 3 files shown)


gazelle/update_repos.go
```diff
@@ -272,5 +272,5 @@ const bzlmodUseRepoNamesUpdMarkerEnd = "    # swift_deps bzlmod use_repo END\n"
 const bzlmodStanzasUpdMarkerStart = "# swift_deps START\n"
 const bzlmodStanzasUpdMarkerEnd = "# swift_deps END\n"
 
-const bzlmodInstructions = `If you have enabled bzlmod, add the following to your 'MODULE.bazel' file to 
+const bzlmodInstructions = `If you have enabled bzlmod, add the following to your 'MODULE.bazel' file to
```
cgrindel, line 275: Was the removal of this space intentional? I think that it will concatenate `toload`.

luispadron: Hmm nope looks like either me or my editor did that, will revert.

luispadron:
Closing in favor of https://github.com/cgrindel/rules_swift_package_manager/pull/366 since it doesnt use a fork and actions wont have credential issues.

PR #366, opened 2023-04-27
fix: local_swift_package relative path resolution

### LINKED ISSUE #363: Local packages do not work with non-root based setups

luispadron:
This fixes the relative path resolution for local_swift_package when the Package.swift is not in the root of the repository.

Closes #363

### DIFF
--- a/gazelle/internal/swift/repo_rule_from_bazel_repo_test.go
+++ b/gazelle/internal/swift/repo_rule_from_bazel_repo_test.go
@@ -13,7 +13,8 @@ import (
 )
 
 func TestRepoRuleFromBazelRepo(t *testing.T) {
-	pkgDir := "/path/to/package"
+	repoDir := "/path/to/repo"
+	pkgDir := "/path/to/repo"
 	diBasename := "swift_deps_index.json"
 
 	t.Run("with pin (source control dep)", func(t *testing.T) {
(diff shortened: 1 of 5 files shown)


luispadron:
Will update the tests in a few

gazelle/internal/swift/repo_rule_from_bazel_repo_test.go
```diff
@@ -58,12 +60,34 @@ func TestRepoRuleFromBazelRepo(t *testing.T) {
...
-		actual, err := swift.RepoRuleFromBazelRepo(br, diBasename, pkgDir)
+		actual, err := swift.RepoRuleFromBazelRepo(br, diBasename, pkgDir, repoDir)
 		assert.NoError(t, err)
 
 		expected := rule.NewRule(swift.LocalSwiftPkgRuleKind, repoName)
 		expected.SetAttr("path", relLocalPkgDir)
 		expected.SetAttr("dependencies_index", fmt.Sprintf("@//:%s", diBasename))
 		assert.Equal(t, expected, actual)
 	})
+	t.Run("without pin (local Swift package with relative path)", func(t *testing.T) {
```
luispadron, line 71: Added this test to assert the fixed behavior

Issue #400, opened 2023-06-05
Bug: Downloading `swift-markdown` fails: `gazelle: dump manifest info for XMLConverter target not found`


luispadron:
When depending on `swift-markdown` `0.2.0` https://github.com/apple/swift-markdown and running the update on packages, it fails with:

```sh
INFO: Elapsed time: 0.398s, Critical Path: 0.11s
INFO: 1 process: 1 internal.
INFO: Running command line: bazel-bin/ThirdParty/SPM/update_packages_to_latest
INFO: Build completed successfully, 1 total action
gazelle: dump manifest info for XMLConverter target not found
```

Can be reproduced by depending on the swift-markdown package in the `examples/ios_app` package.

luispadron:
Not sure where `XMLConverter` stuff is coming from, I'm able to pull the dependencies of this package just fine. The swift-markdown `Package.swift` doesnt look super complicated either so not sure what to go on.

cgrindel:
I am looking at this, right now. So far, there are two issues:

1. The SPM description manifest for swift-markdown has what I call phantom targets. I ran into something similar with products. This is the first time that I have seen phantom targets. I will need to update the code to handle this case.
2. After hacking around #1, I ran into a duplicate Bazel target being generated for https://github.com/apple/swift-cmark/blob/gfm/Package.swift. I am not sure why that is happening. 👀

cgrindel:
@luispadron I have worked through the issues that I mentioned. However, there is one last issue. I believe that `swift-cmark`, a dependency of `swift-markdown` has a bug in their `Package.swift`. The `Package.swift` [explicitly excludes case_fold_switch.inc](https://github.com/apple/swift-cmark/blob/gfm/Package.swift#L56). However, this file is [required by utf8.c](https://github.com/apple/swift-cmark/blob/gfm/src/utf8.c#L237).

Running `swift build` on `swift-cmark` succeeds. This succeeds because clang defaults to looking in the source file's directory when resolving includes. The clang invocation is:

```
[...]

PR #401, opened 2023-06-07
fix: ignore phantom targets and ensure unique target label names

cgrindel:
- The description manifests from SPM can contain undeclared targets (i.e. phantom targets). We ignore these when constructing a Bazel build file.
- It is critical that the Bazel label name for an SPM target not conflict with the label name from a product. Typically, we do this by using the path of the SPM target as the basis for the label name. This PR fixes a bug where a non-unique label name was generated when an SPM target's name and path were equal (i.e., no parent directory in the path).

Related to #400.

### DIFF
--- a/swiftpkg/internal/pkginfo_targets.bzl
+++ b/swiftpkg/internal/pkginfo_targets.bzl
@@ -66,6 +66,18 @@ def _join_path(target, path):
     return _join_path_from_parts(target.path, path)
 
 def _bazel_label_name_from_parts(target_path, target_name):
+    """Create a Bazel label name from a target path and name.
+
+    The logic in this function must stay in sync with BazelLabelFromTarget() in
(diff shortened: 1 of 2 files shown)


swiftpkg/internal/pkginfo_targets.bzl
```diff
@@ -67,15 +67,24 @@ def _join_path(target, path):
...
         name = target_path
     else:
         name = _join_path_from_parts(target_path, target_name)
+
     return name.replace("/", "_")
 
 def _bazel_label_name(target):
     """Returns the name of the Bazel label for the specified target.
 
+    The logic in this function must stay in sync with BazelLabelFromTarget() in
```
luispadron, line 85: I think this comment goes in the other function above

luispadron (approved):
Thank you!

PR #410, opened 2023-06-08
chore: upgrade to rules_go 0.39.1 and remove references to go_sdk

### LINKED ISSUE #408: Migrate to `@rules_go//go`
We would like to remove downstream usages of this repo so that it isn't fetched as often. Is everything in place for you to migrate to `@rules_go//go`?

_Originally posted by @fmeum in https://github.com/bazelbuild/bazel-central-registry/pull/688#discussion_r1223333236_

cgrindel:
- Upgrade `rules_go` to 0.39.1.
- Replace references to `@go_sdk//:bin/go` with `@io_bazel_rules_go//go`.
- Sync'd dependency versions between `WORKSPACE` and `MODULE.bazel`.

Closes #408.

cc: @fmeum

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -71,6 +71,7 @@ gazelle_binary(
 
 # gazelle:prefix github.com/cgrindel/rules_swift_package_manager
 # gazelle:go_naming_convention import
+# gazelle:go_naming_convention_external go_default_library
 gazelle(
     name = "update_build_files",
     command = "fix",
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -98,7 +98,7 @@ execute_binary(
         "mod",
         "tidy",
     ],
-    binary = "@go_sdk//:bin/go",
+    binary = "@io_bazel_rules_go//go",
     execute_in_workspace = True,
 )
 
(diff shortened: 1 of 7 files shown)


BUILD.bazel
```diff
@@ -71,6 +71,7 @@ gazelle_binary(
 
 # gazelle:prefix github.com/cgrindel/rules_swift_package_manager
 # gazelle:go_naming_convention import
+# gazelle:go_naming_convention_external go_default_library
```
fmeum, line 74: Can you say why this line is necessary? It looks like a bug in Gazelle if it is.

cgrindel: If I remove the `gazelle:go_naming_convention_external`, some (not all) of the external dep are stripped of the `:go_default_library`, which results in Bazel analysis errors. TBH, I have not spent much time figuring out what is happening.

One point I think is peculiar is why `:go_default_library` was selected by Gazelle for the external deps when I configured `import` for the local Go targets. I would have thought that it would have inferred that we wanted to use `import` for all of the Go targerts.

fmeum: Yes, that's also what worries me here. With Bzlmod, Gazelle should reliably take the build naming conventions of external repos into account. 

I will take a closer look.

fmeum: I can't reproduce a problem anymore - can you provide me with steps that result in Gazelle generating broken references to external repos?

cgrindel: Created https://github.com/bazelbuild/bazel-gazelle/issues/1557 to track the issue. It includes the repro steps.

Issue #469, opened 2023-07-13
Enable `xcodebuild` test in `ios_sim` example


cgrindel:
The `do_test` for the `ios_sim` example contains commented code to perform a Bazel clean expunge and execute a build via an Xcode project generated by `rules_xcodeproj`. The commented code can be enabled once MobileNativeFoundation/rules_xcodeproj#2355 is closed.

PR #470, opened 2023-07-13
fix: ensure `SDKROOT` is not set when executing SPM commands

cgrindel:
- Ensure that `SDKROOT` is not set when executing SPM commands. The Swift package manifest must be compiled using the host's default SDK (i.e., typically macOS).
- Update `ios_sim` example to demonstrate and test interoperability with `rules_xcodeproj`.

Related to #469.
Related to https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/2355.


cgrindel:
cc: @ brentleyjones @leohidalgo

luispadron (approved):
Great! Thanks for the fix will confirm with our project later today and close out the issue

Issue #378, opened 2023-05-13
Gazelle plugin does not handle swift GRPC / Proto library dependencies correctly


AttilaTheFun:
I'm working on transitioning to use the gazelle plugin provided by this repository. It seems to be working well for my regular code and SPM dependencies, but it fails for swift_proto_library and swift_grpc_library dependencies.

I define Proto targets in BUILD files like:
```
load("@build_bazel_rules_swift//swift:swift.bzl", "swift_proto_library")

swift_proto_library(
    name = "TokenProtos",
    visibility = ["//visibility:public"],
    deps = ["@protos//models/base/tokenmodels:tokenmodels_proto"],
)
```

And GRPC targets like this:
```
load("@build_bazel_rules_swift//swift:swift.bzl", "swift_grpc_library", "swift_proto_library")

swift_proto_library(
    name = "TokenGRPCProtos",
    visibility = ["//visibility:public"],
    deps = ["@protos//services/public/tokenservice:tokenservice_proto"],
)

swift_grpc_library(
    name = "TokenGRPC",
    srcs = ["@protos//services/public/tokenservice:tokenservice_proto"],
    flavor = "client",
    visibility = ["//visibility:public"],
    deps = [":TokenGRPCProtos"],
)
```

Where @protos is a local_repository containing my protos.

I import these from Swift like:
```
import models_base_tokenmodels_tokenmodels_proto
import models_base_usermodels_usermodels_proto
import services_public_tokenservice_tokenservice_proto
import services_public_userservice_userservice_proto
```
Because swift_proto_library and swift_grpc_library generate the module name from the proto package path + target name.

This builds fine with Bazel and inside my generated xcodeproj.

But when I run the Gazelle UpdateBuildFiles target, I get this error:
```
gazelle: Unable to find dependency labels for ant
gazelle: Unable to find dependency labels for os
gazelle: Unable to find dependency labels for models_base_tokenmodels_tokenmodels_proto, models_base_usermodels_usermodels_proto, services_public_tokenservice_tokenservice_proto, services_public_userservice_userservice_proto
```

> Aside: The os system library should maybe be added to the list of known frameworks like Foundation and UIKit. I'm not sure where ant is coming from -- maybe a dependency? Is there a way to query the set of files visible to gazelle?

As far as expected behavior goes, I know adding full support for protos might be challenging but at least it would be nice if it could map those proto imports to manually created swift_proto_library and swift_grpc_library targets so it doesn't complain.

[...]

PR #492, opened 2023-07-23
feat: add initial support for `swift_proto_library` and `swift_grpc_library` to gazelle plugin

AttilaTheFun:
After a bunch of trial and error, I finally got the gazelle plugin to generate swift_proto_library and swift_grpc_library targets similar to the official Go gazelle plugin. rules_swift structures these a little differently than their Go counterparts.

This addresses: https://github.com/cgrindel/rules_swift_package_manager/issues/378

I was successfully able to generate this BUILD.bazel file with:
- proto_library
- go_proto_library
- go_library
- swift_proto_library
- swift_grpc_library (client, client_stubs, and server flavors)

```
load("@rules_swift//swift:swift.bzl", "swift_grpc_library", "swift_proto_library")
load("@rules_proto//proto:defs.bzl", "proto_library")
load("@io_bazel_rules_go//go:def.bzl", "go_library")
load("@io_bazel_rules_go//proto:def.bzl", "go_proto_library")

proto_library(
    name = "tokenservice_proto",
    srcs = ["token_service.proto"],
    deps = ["//models/base/tokenmodels:tokenmodels_proto"],
)

go_proto_library(
    name = "tokenservice_go_proto",
    compilers = ["@io_bazel_rules_go//proto:go_grpc"],
    importpath = "github.com/attilathefun/appengineserver/pb/services/public/tokenservice",
    proto = ":tokenservice_proto",
    deps = ["//models/base/tokenmodels"],
)

go_library(
    name = "tokenservice",
    embed = [":tokenservice_go_proto"],
    importpath = "github.com/attilathefun/appengineserver/pb/services/public/tokenservice",
)

swift_proto_library(
    name = "tokenservice_swift_proto",
    deps = [":tokenservice_proto"],
)

swift_grpc_library(
    name = "tokenservice_swift_grpc",
    srcs = [":tokenservice_proto"],
    flavor = "server",
    deps = [":tokenservice_swift_proto"],
)

swift_grpc_library(
    name = "tokenservice_client_swift_grpc",
    srcs = [":tokenservice_proto"],
    flavor = "client",
    deps = [":tokenservice_swift_proto"],
)

swift_grpc_library(
    name = "tokenservice_client_stubs_swift_grpc",
    srcs = [":tokenservice_proto"],
    flavor = "client_stubs",
    deps = [":tokenservice_client_swift_grpc"],
)

swift_grpc_library(
    name = "tokenservice_server_swift_grpc",
    srcs = [":tokenservice_proto"],
    flavor = "server",
    deps = [":tokenservice_swift_proto"],
)
```

All of this was generated from a single proto file with a GRPC service definition + its request and response types.
The targets build correctly with bazel, though I still need to verify they can be imported correctly from regular swift code.


gazelle/constants.go
```diff
@@ -0,0 +1,12 @@
+package gazelle
+
+const (
+
```
cgrindel, line 4: nit: Can we remove this extra line?

gazelle/generate.go
```diff
@@ -96,6 +125,47 @@ func defaultModuleName(args language.GenerateArgs) string {
...
 
+func generateRuleFromProtoPackage(protoPackageName string, protoPackage proto.Package) []*rule.Rule {
+	swiftProtoLibraryName := strings.TrimSuffix(protoPackageName, "_proto") + swiftProtoSuffix
+	swiftProtoLibrary := rule.NewRule("swift_proto_library", swiftProtoLibraryName)
+	swiftProtoLibrary.SetAttr("deps", []string{":" + protoPackageName})
+	swiftProtoLibrary.SetPrivateAttr(config.GazelleImportsKey, []string{})
+	rules := []*rule.Rule{swiftProtoLibrary}
+
+	if protoPackage.HasServices {
+		// TODO: Should we add a config to selectively generate specific flavors?
```
cgrindel, line 136: If this is something that we are going to address after this PR, could you create an issue and then replace `TODO` with `GHXXX` where `XXX` is the issue number?

gazelle/generate.go
```diff
@@ -96,6 +125,47 @@ func defaultModuleName(args language.GenerateArgs) string {
 	return defaultModuleName
 }
 
+func generateRuleFromProtoPackage(protoPackageName string, protoPackage proto.Package) []*rule.Rule {
+	swiftProtoLibraryName := strings.TrimSuffix(protoPackageName, "_proto") + swiftProtoSuffix
```
cgrindel, line 129: It looks like we use the output of `strings.TrimSuffix(protoPackageName, "_proto")` multiple times in this function. Could you set it to a variable and use it throughout the function?

.bazelrc
```diff
@@ -11,3 +11,8 @@ import %workspace%/ci.bazelrc
 
 # Try to import a local.rc file; typically, written by CI
 try-import %workspace%/local.bazelrc
+
+# The CNIOBoringSSL library uses C++14 features like 'enable_if_t' macro support.
```
cgrindel, line 15: Curious. Which dependency is bringing in CNIOBoringSSL?

cgrindel (changes requested):
This looks good. Thanks for implementing this. 

I left a few minor comments in the code. One thing that we still need to add is an example that exercises these code paths. I suggest we add a new one so clients have an easy-to-find reference.

.bazelrc
AttilaTheFun: This is used by swift_grpc_library from rules_swift here:
https://github.com/bazelbuild/rules_swift/blob/master/third_party/com_github_apple_swift_nio_ssl/BUILD.overlay#L39
Previously you didn't have any swift_grpc_library or swift_proto_library examples so it wasn't being evaluated.

gazelle/generate.go
AttilaTheFun: Ok I created https://github.com/cgrindel/rules_swift_package_manager/issues/509 and I updated the comment to reference it.

AttilaTheFun:
@cgrindel Okay I've got basically everything working except for exposing the GRPC dependencies from rules_swift to targets indexed by rules_swift_package_manager's gazelle plugin.

When updating the build files, I get this message:
```
logan@Logans-MBP grpc_example % bazel run //:update_build_files
INFO: Analyzed target //:update_build_files (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //:update_build_files up-to-date:
  bazel-bin/update_build_files-runner.bash
  bazel-bin/update_build_files
INFO: Elapsed time: 0.437s, Critical Path: 0.32s
INFO: 3 processes: 1 internal, 2 darwin-sandbox.
INFO: Build completed successfully, 3 total actions
INFO: Running command line: bazel-bin/update_build_files
gazelle: Unable to find dependency labels for GRPC, NIOCore, NIOPosix
```

Is there a way to expose external, non-SPM Bazel targets (swift_library, cc_library etc) for indexing to the gazelle plugin?

As an experiment I added grpc-swift as an SPM dependency and it was able to resolve the imports correctly:
```
let package = Package(
    name: "grpc_example",
    dependencies: [
        .package(url: "https://github.com/grpc/grpc-swift", from: "1.16.0"),
    ]
)
```

But then I got linker errors because the modules were redefined between rules_swift and rules_swift_package_manager:
```
bazel-out/darwin_arm64-fastbuild/bin/external/rules_swift_package_manager~override~swift_deps~swiftpkg_swift_nio_ssl/Sources_CNIOBoringSSL.swift.modulemap:1:8: error: redefinition of module 'CNIOBoringSSL'
module "CNIOBoringSSL" {
```

I believe this issue came up once before where rules_swift and some SPM dependency both provided swift-log
Really we *should* be able to resolve these external dependencies through Gazelle, but I'm not sure why that's not working.

examples/grpc_example/MODULE.bazel
```diff
@@ -0,0 +1,66 @@
...
+)
+bazel_dep(
+    name = "gazelle",
+    version = "0.32.0",
+    dev_dependency = True,
+    repo_name = "bazel_gazelle",
+)
+bazel_dep(name = "rules_proto", version = "5.3.0-21.7")
+
+# TODO: Gazelle currently needs protobuf to be named com_google_protobuf.
```
cgrindel, line 51: Let's remove the `TODO: ` part of this comment. I think that the rest of the comment says exactly what is needed.

FYI. I try to avoid `TODO` comments being merged. That is why I prefer that we create GH issues and use `GHXXX` comments instead.

examples/grpc_example/Package.swift
```diff
@@ -0,0 +1,10 @@
+// swift-tools-version: 5.7
+
+import PackageDescription
+
+let package = Package(
+    name: "grpc_example",
+    dependencies: [
+        // .package(url: "https://github.com/grpc/grpc-swift", from: "1.16.0"),
```
cgrindel, line 8: Did you mean to comment this out?

cgrindel (changes requested):

examples/grpc_example/Package.swift
cgrindel: Nvm. I understand why it is commented. out.

cgrindel:
FYI. I pushed a couple of changes to your branch. We do not need the new deps in the parent workspace. Since we do not need those deps, I removed the flag in `.bazelrc`.

cgrindel:
> I believe this issue came up once before where rules_swift and some SPM dependency both provided swift-log
Really we should be able to resolve these external dependencies through Gazelle, but I'm not sure why that's not working.

We sure did. 😞

I think that we should move the GRPC stuff out of `rules_swift` and then have this new Bazel module leverage `rules_swift_package_manager` to pull down the external dependencies. This would remove the external deps from `rules_swift` and allow clients to select the version of the external dependencies that they want. 

The alternative is to shoehorn support for these special packages into the Gazelle plugin. The problem is that it will never be foolproof. For instance, if a client wants to use the GRPC functionality and wants to use `swift-log` or any of the other deps declared in `rules_swift`, they need to use the versions defined in the rules. If they added it to their `Package.swift`, they will receive the redefinition of module errors.

cgrindel:
I just posted an [idea on how we can manage this on Slack](https://bazelbuild.slack.com/archives/CD3QY5C2X/p1692055426375909).

examples/grpc_example/Package.swift
cgrindel, line 8: Let's remove this commented code.

cgrindel (changes requested):
LGTM except we need to remove some commented code. Once that is done, we should be good to merge.

examples/grpc_example/Package.swift
cgrindel: I went ahead and made the change.

cgrindel:
I ran `//:tidy_all` and updated this branch. I think that we are good to go.

cgrindel (approved):

Issue #504, opened 2023-07-28
Package with Two Products, with an Intradependency from One Product


ekohlwey:
Hi there,
I'm trying to use the swift Porter Stemmer library from https://github.com/scaraux/Swift-Porter-Stemmer-2.

During the build for `@swiftpkg_swift_porter_stemmer_2//:PorterStemmer2_Classes_Swift_PorterStemmer2`, there is a compiler error:

```
ERROR: /private/var/tmp/_bazel_ekohlwey/80429706e7731d7db50de017c57ccb50/external/swiftpkg_swift_porter_stemmer_2/BUILD.bazel:136:14: Compiling Swift module @swiftpkg_swift_porter_stemmer_2//:PorterStemmer2_Classes_Swift_PorterStemmer2 failed: (Exit 1): worker failed: error executing command (from target @swiftpkg_swift_porter_stemmer_2//:PorterStemmer2_Classes_Swift_PorterStemmer2) bazel-out/darwin_arm64-opt-exec-2B5CBBC6/bin/external/build_bazel_rules_swift/tools/worker/worker swiftc ... (remaining 1 argument skipped)
error: emit-module command failed with exit code 1 (use -v to see invocation)
external/swiftpkg_swift_porter_stemmer_2/PorterStemmer2/Classes/Swift/PorterStemmer.swift:17:8: error: no such module 'libstemmer'
import libstemmer
       ^
external/swiftpkg_swift_porter_stemmer_2/PorterStemmer2/Classes/Swift/PorterStemmer.swift:17:8: error: no such module 'libstemmer'
import libstemmer
       ^
```

I take it that this is because Bazel is unaware of the dependency between the `PorterStemmer2` product and the `libstemmer` product.

cgrindel:
Is there any chance that you can put up a repro? It will help me be able to debug this quicker.

cgrindel:
[Update from repro](https://github.com/cgrindel/rules_swift_package_manager/pull/505#issuecomment-1659239418).

It turns out that the Package.swift for the 0.1.1 version of Swift-Porter-Stemmer-2 has a bug. The PortStemmer2 module should have a dependency on libstemmer. Luckily, someone has already submitted [a fix](https://github.com/scaraux/Swift-Porter-Stemmer-2/commit/5e90cdbc3700a1bf0b8083ba1117c8c4e48669a6). So, you just need to use the repo at that commit.

Let me know if this fixes the issue on your end.

[...]

PR #505, opened 2023-07-28
chore: reproduce #504, add `Swift-Porter-Stemmer-2` to `interesting_deps` example

ekohlwey:
This reproduces the error reported in #504. The index updates are the result of running `bazel :tidy` on the example project.

### DIFF
--- a/examples/interesting_deps/BUILD.bazel
+++ b/examples/interesting_deps/BUILD.bazel
@@ -43,7 +43,8 @@ swift_binary(
     deps = [
         "@swiftpkg_libwebp_xcode//:libwebp",
         "@swiftpkg_swift_log//:Sources_Logging",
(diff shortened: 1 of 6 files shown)


cgrindel:
🙏 Thanks for putting this together. I will take a look.

cgrindel:
@ekohlwey It turns out that the `Package.swift` for the 0.1.1 version of `Swift-Porter-Stemmer-2` has a bug. The `PortStemmer2` module should have a dependency on `libstemmer`. Luckily, someone has already submitted [a fix](https://github.com/scaraux/Swift-Porter-Stemmer-2/commit/5e90cdbc3700a1bf0b8083ba1117c8c4e48669a6). So, you just need to use the repo at that commit. I updated the repro to use the latest commit. I also added `guard` statement to avoid unwrap errors.

Let me know if this fixes the issue on your end.

cgrindel:
Closing the PR as I believe the resolution for the issue is described in #504.

PR #507, opened 2023-08-10
fix: check description JSON and dump JSON for resources

### LINKED ISSUE #491: Packages using the `Bundle.module` SwiftPM-generated bundle accessor fails to compile
SwiftPM generates a `Bundle.module` static Swift accessor to access the package's resource bundle.

If such a package is integrated with rules_swift_package_manager it fails to compile because rules_swift_package_manager/bazel doesn't codegen that same accessor.

For example pulling in the Stripe SDK:

```swift
.package(
  url: "https://github.com/stripe/stripe-ios",
  from: "23.11.0"
)
```

Results in a build failure:

```
external/rules_swift_package_manager~0.7.1~swift_deps~swiftpkg_stripe_ios/StripeCore/StripeCore/Source/Helpers/StripeCoreBundleLocator.swift:15:48: error: type 'Bundle' has no member 'module'
        static let spmResourcesBundle = Bundle.module
                                        ~~~~~~ ^~~~~~
```

Separate but related: I'm curious how folks generally migrate code that uses `Bundle.module` from SwiftPM to Bazel for code that they control (e.g. not external dependencies).

cgrindel:
When looking for resources in an SPM target, we only looked in the dump JSON. We need to check the description JSON too. I am not sure why the dump JSON does not list the resources.

Also added `stripe_example` to demonstrate that #491 is fixed.

Closes #491.


cgrindel:
@luispadron @jpsim There are many files in this PR. Most are [the files that I brought over](https://github.com/stripe/stripe-ios/tree/master/Example) for the `stripe_example`. The fix is in the `swiftpkg/internal/pkginfos.bzl` file. Please focus your review on the files outside of the example.

cgrindel:
@jpsim Thanks. A repro in the `strip_example` would be useful. I will go ahead and merge this PR.

luispadron (commented):
Belated lgtm thanks!

Issue #553, opened 2023-08-31
swift_update_packages does not respect `swift_deps_index` with bzlmod


luispadron:
I have the following `swift_update_packages` declaration in the `ThirdParty/SPM/BUILD.bazel` file:

```python
swift_update_packages(
    name = "update_packages",
    gazelle = ":rules_swift_package_manager_gazelle",
    package_manifest = "ThirdParty/SPM/Package.swift",
    swift_deps_index = "ThirdParty/SPM/swift_packages_index.json",
    update_bzlmod_use_repo_names = True,
    update_bzlmod_stanzas = True,
    generate_swift_deps_for_workspace = False,
)
```

When I `bazel run //ThirdParty/SPM:update_packages` the `MODULE.bazel` is updated accordingly:

```python
# swift_deps START
swift_deps = use_extension(
    "@rules_swift_package_manager//:extensions.bzl",
    "swift_deps",
)
swift_deps.from_file(
    deps_index = "//:swift_deps_index.json",
)
use_repo(
    swift_deps,
    ... # snip
)
# swift_deps END
```

It should instead look like:

```python
# swift_deps START
swift_deps = use_extension(
    "@rules_swift_package_manager//:extensions.bzl",
    "swift_deps",
)
swift_deps.from_file(
    deps_index = "//ThirdParty/SPM:swift_packages_index.json",
)
use_repo(
    swift_deps,
    ... # snip
)
# swift_deps END
```

luispadron:
Going to try and fix this up this week

luispadron:
So im seeing the Vapor example should run into this issue but it doesnt because of where `#swift_deps` comment is:

Currently it's:

```python
swift_deps = use_extension(
    "@rules_swift_package_manager//:extensions.bzl",
    "swift_deps",
)
swift_deps.from_file(
    deps_index = "//swift:deps_index.json",
)
use_repo(
    swift_deps,
    # swift_deps START
    "swiftpkg_fluent",
    "swiftpkg_fluent_sqlite_driver",
    "swiftpkg_vapor",
    # swift_deps END
)
```

If I remove that entire block and run `//swift:update_pkgs` then it turns into the bug above:

```python
# swift_deps START
swift_deps = use_extension(
    "@rules_swift_package_manager//:extensions.bzl",
    "swift_deps",
)
swift_deps.from_file(
    deps_index = "//:swift_deps_index.json",
)
use_repo(
    swift_deps,
[...]

PR #568, opened 2023-09-06
fix: bzlmod deps_index custom path

luispadron:
Fixes https://github.com/cgrindel/rules_swift_package_manager/issues/553 
The stanza template should account for deps index path.

### DIFF
--- a/gazelle/internal/swift/bzlmod_test.go
+++ b/gazelle/internal/swift/bzlmod_test.go
@@ -30,7 +30,7 @@ func TestBzlmodStanzas(t *testing.T) {
 	di.AddPackage(awesomePkg, anotherPkg)
 	di.AddDirectDependency(awesomeRepoId, anotherRepoID)
 
-	actual, err := swift.BzlmodStanzas(di)
+	actual, err := swift.BzlmodStanzas(di, "path/to/root", "path/to/root", "path/to/root/swift_deps_index.json")
 	assert.NoError(t, err)
 	expected := `swift_deps = use_extension(
     "@rules_swift_package_manager//:extensions.bzl",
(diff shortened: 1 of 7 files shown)


examples/vapor_example/BUILD.bazel
```diff
@@ -23,7 +23,7 @@ gazelle_binary(
 gazelle(
     name = "update_build_files",
     args = [
-        "-swift_dependency_index=swift/deps_index.json",
+        "-swift_dependency_index=swift/swift_deps_index.json",
```
cgrindel, line 26: Do we need to rename the file to be `swift_deps_index.json`? The original intent was to allow the client to place the file where they wanted with whatever name they wanted.

gazelle/internal/swift/bzlmod_test.go
```diff
@@ -47,3 +47,44 @@ use_repo(
...
+		Local: &swift.LocalPackage{
+			Path: "path/to/another",
+		},
+	}
+
+	di := swift.NewDependencyIndex()
+	di.AddPackage(awesomePkg, anotherPkg)
+	di.AddDirectDependency(awesomeRepoId, anotherRepoID)
+
+	actual, err := swift.BzlmodStanzas(di, "path/to/root", "path/to/root/swift", "path/to/root/swift/swift_deps_index.json")
```
cgrindel, line 74: Can this test not use the default name `swift_deps_index.json` (e.g., `deps_index.json`)?

examples/vapor_example/BUILD.bazel
luispadron: ah I see the default is `swift_deps_index`, i thought i was changing the name to validate both the name being custom and the path but i'll revert this to `deps_index` since thats custom

gazelle/internal/swift/bzlmod_test.go
luispadron: Updated!

cgrindel (approved):

PR #576, opened 2023-09-07
fix: auto discover select resources in external Swift packages

cgrindel:
[SPM discovers select resource types](https://github.com/apple/swift-package-manager/blob/main/Documentation/PackageDescription.md#resource) if they exist under a target. This change looks for the following resources and adds them as a resource bundle to the target:

```python
XIB = ["nib", "xib", "storyboard"]
ASSET_CATALOG = ["xcassets"]
STRING_CATALOG = ["xcstrings"]
COREDATA = ["xcdatamodeld", "xcdatamodel", "xcmappingmodel"]
METAL = ["metal"]
```

In addition, add `messagekit_example`. The `MessageKit` package contains targets with some of these auto-discoverable resources.

Closes #491.

PR #577, opened 2023-09-07
chore: refactor how resources are specified on `apple_resource_bundle`

### LINKED ISSUE #575: Refactor `swiftpkg_build_files.apple_resource_bundle` to separate glob patterns from file paths
The current implementation creates `glob()` for all of the resources with the include list being patterns and file paths. We should only create the glob if we have patterns and the file paths should be a simple list.

cgrindel:
It turns out that the auto-discovered Swift package resources (#576) are listed in the description JSON. This PR ended up removing the auto-discovery code and having each resource listed as a file path.

Updated the `testutils` to support stub calls to `repository_files.is_directory()`.

Closes #575.

### DIFF
--- a/swiftpkg/internal/pkginfos.bzl
+++ b/swiftpkg/internal/pkginfos.bzl
@@ -241,13 +241,37 @@ def _new_target_from_json_maps(
     linker_settings = _new_linker_settings_from_dump_json_list(
         dump_map["settings"],
     )
-    resources_map = dump_map.get("resources", [])
-    if len(resources_map) == 0:
-        resources_map = desc_map.get("resources", [])
+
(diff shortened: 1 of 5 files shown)


luispadron (approved):

cgrindel:
@luispadron One of the examples was failing. While I was investigating, I realized that the description JSON contains auto-discovered resources. So, I removed the code that searched for auto-discovered resources. 🤞

cgrindel:
> While I was investigating, I realized that the description JSON contains auto-discovered resources. So, I removed the code that searched for auto-discovered resources. 

Well, it looks like SPM does not list all of the resources. So,I am going to add back the auto-discovery and dedupe the results with what SPM provides.

Issue #510, opened 2023-08-14
Packages overriding `cxxLanguageStandard` fail to compile


jpsim:
For example, Sentry defines `cxxLanguageStandard: .cxx14`:

https://github.com/getsentry/sentry-cocoa/blob/0d32275107b8a0160580f99b52ddbcee94196ddb/Package.swift#L59

But when compiling it via rules_swift_package_manager, clang is passed `-std=gnu+11`, causing the build to fail because the package requires C++ 14.

cgrindel:
I have not been sure how we want to solve this. It is possible to add these flags to the generated targets. However, that could lead to weird issues if conflicting flags are specified elsewhere. In general, I believe the recommended advice for specifying these types of flags is to add them to the `.bazelrc` so that all targets are built in a compatible manner. That is what I have been doing for the [examples that require these flags](https://github.com/cgrindel/rules_swift_package_manager/blob/main/examples/vapor_example/.bazelrc#L10-L13). 

What would you want the ruleset to do? The following are some options that have been discussed in the past.

1. When `cxxLanguageStandard` features are used in a module, report this to the client when they run `//:swift_update_pkgs`.
2. Provide a configuration option to add flags to the relevant targets.
3. Something else?

Right now, I am inclined to implement option 1. Right now, the client experience is not ideal, having them trip over the compilation errors. We can even show the suggested `.bazelrc` line(s).

Let me know what think.

jpsim:
I finally figured out why some packages were respecting `build --cxxopt='-std=c++14'` but Sentry doesn't.

It's because Sentry uses Objective-C++. Do you know if there's an equivalent ObjC++ flag for `--cxxopt`? I tried with `--objccopt="-ObjC++" --objccopt="-std=c++14"` and Bazel wasn't happy with that because it also appears to apply to normal ObjC files:

```
<unknown>:0: warning: argument unused during compilation: '-ObjC++'
<unknown>:0: error: invalid argument '-std=c++14' not allowed with 'Objective-C'
<unknown>:0: error: clang importer creation failed
<unknown>:0: warning: argument unused during compilation: '-ObjC++'
<unknown>:0: error: invalid argument '-std=c++14' not allowed with 'Objective-C'
<unknown>:0: error: clang importer creation failed
```

Here's a minimal repro, adding sentry-cocoa to one of the existing examples: https://github.com/cgrindel/rules_swift_package_manager/compare/main...jpsim:swift_bazel:jp/sentry_example

[...]

PR #609, opened 2023-09-22
fix: allow .cpp when listing C++ files

Thooms:
Fixes the issue raised in https://github.com/cgrindel/rules_swift_package_manager/issues/510

In the specific case of Sentry (https://github.com/getsentry/sentry-cocoa), the generated `objc_library` target for the main `Sentry` module included the right headers, allowing it to build, but not the right sources (for instance https://github.com/getsentry/sentry-cocoa/blob/72c8d84c86cebd726dd3b34ab9ff09a3596dd5f2/Sources/Sentry/SentrySamplingProfiler.cpp#L169). This cause the build to work but not the link.


jpsim (approved):
Good catch!

Thooms:
Sorry @jpsim I'm having to update with `main`, can I let you merge it whenever you see fit?

cgrindel (approved):

PR #614, opened 2023-09-25
dirty hack for shake_ios integration

joune:


cgrindel:
The fix for this is in #615.

PR #629, opened 2023-09-30
fix: support `#import` with whitespace between `#` and `import`

cgrindel:
The [Sentry package](https://github.com/getsentry/sentry-cocoa) uses [an unusual import syntax for the SystemConfiguration framework](https://github.com/getsentry/sentry-cocoa/blob/e32423038e99f0c5d868744421d664cdd55c7383/Sources/Sentry/include/SentryReachability.h#L31):

```objc
#    import <SystemConfiguration/SystemConfiguration.h>
```

This PR updates the parsing logic for `#import` lines to support whitespace between the `#` and the `import`.

Related to #510.


cgrindel:
cc: @Thooms @jpsim

Thooms:
Awesome!

luispadron (approved):
That is odd. Thanks for making it work!

jpsim:
“unusual“ is right, good find!

PR #631, opened 2023-09-30
fix: Handle Xcode 15 swift package format

AttilaTheFun:
In Xcode 15 / Swift 5.9 Apple changed the output format of `swift package dump-package` which rules_swift_package_manager parsed. This caused the JSON deserialization to fail.

Specifically, the SourceControlLocation type changed from a single RemoteLocation to an array of RemoteLocation(s) and these are now objects with the key "urlString" instead of "url".

I added support for both the Xcode 14 / Swift 5.8 format as well as the new Xcode 15 / Swift 5.9 format.
Whenever we drop support for Xcode 14, the support for the old format can be removed.

I verified the change from my repo by using a local_path_override with this patch and I verified that before swift_update_packages failed with the error:
```
gazelle: index 0 expected to be type 'string', but was type 'map[string]interface {}'
```
While now it succeeds.

### DIFF
--- a/gazelle/internal/spdump/dependency.go
+++ b/gazelle/internal/spdump/dependency.go
@@ -60,11 +60,11 @@ func (sc *SourceControl) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
-type sourceControlLocationXcode15 struct {
+type sourceControlLocationSwift59 struct {
 	Remote []*RemoteLocation
 }
 
-type sourceControlLocationXcode14 struct {
+type sourceControlLocationSwift58 struct {
 	Remote *RemoteLocation
 }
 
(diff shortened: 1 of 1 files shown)


cgrindel:
Did you happen to check if this structure is used in the Starlark code? We parse dump and description JSON in `pkginfos.bzl`, as well.

gazelle/internal/spdump/dependency.go
```diff
@@ -59,19 +60,67 @@ func (sc *SourceControl) UnmarshalJSON(b []byte) error {
...
+
 // A SourceControlLocation represents the location of a source control repository.
 type SourceControlLocation struct {
 	Remote *RemoteLocation
 }
 
+func (scl *SourceControlLocation) UnmarshalJSON(b []byte) error {
+
+	// Try the Xcode 15 format first:
+	var sclx15 sourceControlLocationXcode15
```
cgrindel, line 79: Instead of using Xcode versions, could we use Swift versions? That will make it clearer for anyone working on non-Apple systems.

cgrindel (commented):
Overall, LGTM. I added a comment about the naming convention is used. Also, it looks like we [parse the description JSON in the pkginfos.bzl](https://github.com/cgrindel/rules_swift_package_manager/blob/main/swiftpkg/internal/pkginfos.bzl#L439-L442). Could you check if that changed?

AttilaTheFun:
@cgrindel I'm not sure how to test the pkginfos.bzl file but it doesn't appear to parse the location / remote / urlString so it's probably still working.

gazelle/internal/spdump/dependency.go
AttilaTheFun: Ok I changed the names to refer to Swift 5.9 / 5.8 instead.

cgrindel (approved):

PR #632, opened 2023-09-30
feat: Add directive to disable manual tag

AttilaTheFun:
This PR introduces a directive to disable the addition of the manual tag to generated swift_library targets.

I understand the intent behind the tag and am ok with keeping it the default. But for more advanced users who understand that they need to pass the apple platform type configuration, it is useful to be able to build globs of these targets without a dedicated build test target.

### DIFF
--- a/examples/pkg_manifest_minimal/BUILD.bazel
+++ b/examples/pkg_manifest_minimal/BUILD.bazel
@@ -14,6 +14,7 @@ tidy(
 
 # Ignore the Swift build folder
 # gazelle:exclude .build
+# gazelle:swift_library_tags -
 
 gazelle_binary(
     name = "gazelle_bin",
--- a/examples/pkg_manifest_minimal/BUILD.bazel
+++ b/examples/pkg_manifest_minimal/BUILD.bazel
@@ -14,6 +14,8 @@ tidy(
 
 # Ignore the Swift build folder
 # gazelle:exclude .build
+
+# Omit tags = ["manual"] from the generated swift library targets
 # gazelle:swift_library_tags -
 
 gazelle_binary(
--- a/gazelle/internal/swift/rules_common.go
+++ b/gazelle/internal/swift/rules_common.go
@@ -14,16 +14,20 @@ func rulesForLibraryModule(
 	srcs []string,
 	swiftImports []string,
 	shouldSetVis bool,
+	swiftLibraryManualTagDisabled bool,
 	buildFile *rule.File,
 ) []*rule.Rule {
 	name, moduleName := ruleNameAndModuleName(buildFile, LibraryRuleKind, defaultName, defaultModuleName)
 	r := rule.NewRule(LibraryRuleKind, name)
 	setCommonSwiftAttrs(r, moduleName, srcs, swiftImports)
 	setVisibilityAttr(r, shouldSetVis, []string{"//visibility:public"})
-	// Mark swift_library targets as manual. We do this so that they are always
-	// built from a leaf node which can provide critical configuration
-	// information.
-	r.SetAttr("tags", []string{"manual"})
+	if !swiftLibraryManualTagDisabled {
+		// Mark swift_library targets as manual. We do this so that they are always
+		// built from a leaf node which can provide critical configuration
+		// information.
+		r.SetAttr("tags", []string{"manual"})
+	}
+
 	return []*rule.Rule{r}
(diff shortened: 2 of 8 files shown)


gazelle/internal/swiftcfg/swift_config.go
```diff
@@ -43,8 +43,13 @@ type SwiftConfig struct {
 
 	GenerateSwiftDepsForWorkspace bool
 
+	// The naming convention to apply to the module names derived from the directory names.
+	// The default behavior uses the name verbatim while PascalCase will convert snake_case to PascalCase.
 	ModuleNamingConvention string
 
+	// Whether the default behavior to tag generated swift library targets as manual should be disabled.
+	SwiftLibraryManualTagDisabled bool
```
luispadron, line 51: I wonder if we could instead allow configuring the tags that get added. Using none, you'd get this behavior but it's extensible to custom tags too. 

Not super familiar with what's allow in these tags though

luispadron (approved):

luispadron (commented):
I'd also consider adding an example that tests this behavior

cgrindel:
> I'd also consider adding an example that tests this behavior

I agree that we should have an example use this tag. How about we update the [pkg_manifest_minimal example](https://github.com/cgrindel/rules_swift_package_manager/tree/main/examples/pkg_manifest_minimal) to use it. I believe it is a non-Apple, Swift example.

gazelle/internal/swiftcfg/swift_config.go
cgrindel: A client should not need to add the tag. In fact, I would seriously discourage it for most situations. Do you think this will be confusing for clients who see it in the doc?

AttilaTheFun: @cgrindel @luispadron I kinda like the idea of giving users an option to specify a list of tags to apply to generated targets, with "manual" being the default. I could imagine someone might want to add a "generated" tag or something like that.

examples/pkg_manifest_minimal/BUILD.bazel
```diff
@@ -14,6 +14,7 @@ tidy(
 
 # Ignore the Swift build folder
 # gazelle:exclude .build
+# gazelle:swift_library_tags -
```
cgrindel, line 17: Please add a comment above the directive explaining what it is doing.

gazelle/internal/swift/rules_common.go
```diff
@@ -14,16 +14,17 @@ func rulesForLibraryModule(
...
+	swiftLibraryTags []string,
 	buildFile *rule.File,
 ) []*rule.Rule {
 	name, moduleName := ruleNameAndModuleName(buildFile, LibraryRuleKind, defaultName, defaultModuleName)
 	r := rule.NewRule(LibraryRuleKind, name)
 	setCommonSwiftAttrs(r, moduleName, srcs, swiftImports)
 	setVisibilityAttr(r, shouldSetVis, []string{"//visibility:public"})
-	// Mark swift_library targets as manual. We do this so that they are always
-	// built from a leaf node which can provide critical configuration
-	// information.
```
cgrindel, line 25: Can you put this comment by the `swiftLibraryTags` code in `config.go`? I want to ensure that anyone looking at the code understands why `manual` tags are recommended.

cgrindel (commented):
Nice change to the comma-separated list of tags. I have just a couple of small comments. We should be good to go after that.

AttilaTheFun:
Ok I think I addressed all of your comments @cgrindel
Please let me know if I should change anything else

cgrindel (approved):

gazelle/internal/swiftcfg/swift_config.go
```diff
@@ -43,8 +43,14 @@ type SwiftConfig struct {
 
 	GenerateSwiftDepsForWorkspace bool
 
+	// The naming convention to apply to the module names derived from the directory names.
+	// The default behavior uses the name verbatim while PascalCase will convert snake_case to PascalCase.
 	ModuleNamingConvention string
 
+	// The set of tags to apply to generated swift library targets.
+	// Defailts to ["manual"]
```
luispadron, line 51:
```suggestion
	// Defaults to ["manual"]
```

luispadron:
Sweet thanks for updating!

PR #639, opened 2023-10-03
chore(perf): refactor checking path for directory

### LINKED ISSUE #582: Refactor `repository_files.is_directory()`
It should be as simple as `repostiory_files.path_exists()`

cgrindel:
- Instead of shelling out to test each path, the new code finds all of the directories under the target in one shell call. The result is then added to a set and the set is checked.
  - The starlark profile for the `soto_example` dropped from 350 seconds to 46 seconds.
- Simplified the implementation for `repository_files.is_directory()`.

Closes #582.

### DIFF
--- a/swiftpkg/internal/pkginfos.bzl
+++ b/swiftpkg/internal/pkginfos.bzl
@@ -780,12 +780,19 @@ def _new_swift_src_info_from_sources(repository_ctx, target_path, sources):
         target_path,
     )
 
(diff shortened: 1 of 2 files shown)


cgrindel:
cc: @jpsim

luispadron (approved):
Nice!

jpsim (approved):
🙌

Issue #509, opened 2023-08-12
feat: Add configuration to gazelle plugin to selectively generate swift grpc library flavors.


AttilaTheFun:
Initially I just made the gazelle plugin generate all of the flavors (client, client_stubs, and server). This shouldn't cause any issues since, if nothing depends on them, they'll never be evaluated. To produce "cleaner" BUILD files, though, it would be nice to have a gazelle configuration to only generate a subset of these. E.g. if the protos are only being used on the client, there is no need to generate the server protos.

cgrindel:
Per [a recent Slack thread](https://bazelbuild.slack.com/archives/CD3QY5C2X/p1696950765107109?thread_ts=1696946243.388449&cid=CD3QY5C2X), we need to provide a way for a client to disable the generation of all Swift GRPC targets.

PR #650, opened 2023-10-09
feat: Add directive to control set of GRPC flavors generated by gazelle plugin

AttilaTheFun:
This PR addresses #509 by introducing a new directive to control the set of GRPC flavors for which the gazelle plugin should generate targets. Currently it generates all three (client, client_stubs and server) but this is probably unnecessary for most people. In my case, I'm using Go on the server and Swift on the client and I don't use the client stubs so I really only need the client targets.

To test, I tried deleting the grpc_example generated BUILD file and regenerating it with just client,server and verified it only included those targets. I then added back client_stubs and verified it added that too. (That's why the order of the targets is now swapped in the BUILD file.)


gazelle/config.go
```diff
@@ -147,11 +149,13 @@ func (*swiftLang) Configure(c *config.Config, rel string, f *rule.File) {
 		switch d.Key {
 		case moduleNamingConventionDirective:
 			if d.Value == swiftcfg.PascalCaseModuleNamingConvention {
+				log.Println("module naming convention ", swiftcfg.PascalCaseModuleNamingConvention)
```
cgrindel, line 152: Is this necessary? In the future, it would be nice to have a verbose mode that told us this kind of information. I am unsure that we should just print it out by default.

gazelle/config.go
```diff
@@ -147,11 +149,13 @@ func (*swiftLang) Configure(c *config.Config, rel string, f *rule.File) {
...
 				sc.ModuleNamingConvention = swiftcfg.PascalCaseModuleNamingConvention
 			} else {
+				log.Println("module naming convention ", swiftcfg.MatchCaseModuleNamingConvention)
```
cgrindel, line 155: Is this necessary?

cgrindel (commented):
Overall, LGTM. Just a couple of questions about some print statements.

gazelle/config.go
AttilaTheFun: No sorry I was debugging something and forgot to take these out. I'll remove them.

AttilaTheFun:
@cgrindel okay I cleaned up the print statements and also added a second directive to control the proto library generation.

Now if you add:
```
# gazelle:swift_generate_proto_libraries false
# gazelle:swift_generate_grpc_libraries_with_flavors -
```
To your BUILD file, no swift proto libraries or grpc libraries will be generated.
I verified this by deleting the generated BUILD files in the grpc_example and regenerating them with those directives.
They only contained the proto_library from the main gazelle proto plugin.

I then added back:
```
# gazelle:swift_generate_proto_libraries true
# gazelle:swift_generate_grpc_libraries_with_flavors client,client_stubs,server
```
And verified it regenerated the targets as expected.

This is also the default behavior to preserve the existing functionality for any users who might have started relying on this already.

cgrindel (approved):
LGTM

cgrindel:
@AttilaTheFun It looks like the GRPC examples are failing and we may need to run `//:tidy` on the repo.

PR #660, opened 2023-10-16
feat: add support for `.enableExperimentalFeature()`

jpsim:
This fixes the `swift-atomics` package.

### DIFF
--- a/swiftpkg/internal/pkginfos.bzl
+++ b/swiftpkg/internal/pkginfos.bzl
@@ -1212,19 +1212,25 @@ def _new_swift_settings(build_settings):
     """
     defines = []
     unsafe_flags = []
+    experimental_features = []
     for bs in build_settings:
         if bs.kind == build_setting_kinds.define:
             defines.append(bs)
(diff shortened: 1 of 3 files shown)


cgrindel:
Is there something that we can add to the examples that will exercise this?

cgrindel:
Is this PR to support [this flag](https://github.com/apple/swift-atomics/blob/cd142fd2f64be2100422d658e7411e39489da985/Package.swift#L25)?

cgrindel:
Instead of adding it to an example, perhaps we can add a test in `swiftpkg_build_files_tests.bzl`.

jpsim:
> Is this PR to support [this flag](https://github.com/apple/swift-atomics/blob/cd142fd2f64be2100422d658e7411e39489da985/Package.swift#L25)?

Yes

> Instead of adding it to an example, perhaps we can add a test in swiftpkg_build_files_tests.bzl.

I'll take a look.

cgrindel (approved):

PR #702, opened 2023-10-30
chore: Add tca_example to CI tests

### LINKED ISSUE #643: Fix or remove `tca_example`
The example exists, but was not added to `examples_infos`. Hence, it is not tested as part of CI.

luispadron:
Fixes issue #643

Other changes:

- Updates the `tca_example` so that it builds and tests. 
- I removed the case studies example, it's pretty large and when I tried updating it to the latest example there were a few errors. In the long term it might become tough to maintain that many example files, and so I added a simplified App which includes TCA in both sources and tests.
  - The example is based off their README


cgrindel (approved):
Sweet! Thanks.

luispadron:
Weird tests are failing CI with:

```
2023-10-30 22:04:47,509 Running test-without-building with device D1410AAB-B94A-4BBE-BD3D-6B1691694C1D
2023-10-30 22:07:20,223 The xcodebuild command got stuck and has not started test in 150. Will kill the command directly.
2023-10-30 22:07:29,576 xcodebuild command can not launch test on device/simulator in 150s.
```

These pass locally though... looking to see why

luispadron:
It was due to a timeout with setting the app host, thats not important for what were testing so i removed that and the tests now pass quickly.

PR #710, opened 2023-10-31
fix: do not generate a modulemap for Objc targets, if they provide one

### LINKED ISSUE #712: Allow Swift packages to be built outside of the sandbox
The default `--spawn_strategy` for Bazel is `remote,worker,sandboxed,local`. Some rulesets (e.g. [rules_xcodeproj](https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/6c186331c82f3cbc82e2e7fdfacb4873e409e094/xcodeproj/internal/templates/xcodeproj.bazelrc#L66-L68)) suggest building with a spawn strategy value of `remote,worker,local` because sandboxing on MacOS is slow. Unfortunately, some Swift packages (e.g., `firebase-ios-sdk`) do not successfully build outside of the sandbox. It would be ideal if `rules_swift_package_manager` generated a `BUILD.bazel` file that built successfully for both `sandboxed` and `local` spawn strategies.

### Related Links

- https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/2703.
- [Slack thread discussing issue with Brently and thi](https://bazelbuild.slack.com/archives/C0429BN0F8V/p1698710034083049)
- [Slack thread summarizing workaround](https://bazelbuild.slack.com/archives/C05NZSHE7FU/p1698853000609299)
- [Bazel spawn strategies](https://bazel.build/docs/user-manual)
- [Bazel default spawn strategy list](https://bazel.build/reference/command-line-reference#flag--spawn_strategy)

cgrindel:
The goal of this PR is to selectively generate modulemap files for ObjC targets. If an ObjC target provides a modulemap, we will not generate a modulemap. This avoids `redefinition of module` errors when running outside of the Bazel action sandbox. While we may not generate modulemaps for some ObjC targets, we must continue to create the `generate_modulemap` targets. The logic that assigns dependencies has no way to know whether a modulemap target has been generated (i.e., the ObjC target may be a dependency in another Swift package). Hence, `generate_modulemap` has been modified to have a noop mode.

- Add `noop` attribute to `generate_modulemap` rule. If `True`, a modulemap is generated and the appropriate providers are populated. If `False, a modulemap is not generated and empty providers are returned.
- Update `clang_files.collect_files()` to return the path of the modulemap file, if one is found.
- Update the build file generation to set the `noop` attribute for `generate_modulemap` based upon whether a modulemap file was found for the module.
- Update `firebase_example` to use `rules_xcodeproj` to generate an Xcode project and ensure that it builds using the `rules_xcodeproj` recommended build API>

Related to https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/2703.
Closes #712.

### DIFF
--- a/swiftpkg/internal/generate_modulemap.bzl
+++ b/swiftpkg/internal/generate_modulemap.bzl
@@ -19,8 +19,9 @@ ModuleMapInfo = provider(
 def _generate_modulemap_impl(ctx):
     module_name = ctx.attr.module_name
 
-    # Decide whether we should generate the modulemap.
-    if not ctx.attr.public:
+    # Decide whether we should generate the modulemap. If we do not generate
+    # the modulemap, we still need to return expected providers.
+    if ctx.attr.noop:
         return [
             DefaultInfo(files = depset([])),
             ModuleMapInfo(
@@ -43,12 +44,6 @@ def _generate_modulemap_impl(ctx):
         for dep in ctx.attr.deps
         if ModuleMapInfo in dep
     ]
-
-    # out_filename = "{}/module.private.modulemap".format(module_name)
-    # out_filename_template = "{}/module.modulemap"
-    # if not ctx.attr.public:
-    #     out_filename_template = "{}/module.private.modulemap"
-    # out_filename = out_filename_template.format(module_name)
 
     out_filename = "{}/module.modulemap".format(module_name)
     modulemap_file = ctx.actions.declare_file(out_filename)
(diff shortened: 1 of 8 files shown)


cgrindel:
@Brentley @thii If you have a moment, I would like to get your thoughts on my solution for the `redefinition of module` errors.

swiftpkg/internal/generate_modulemap.bzl
```diff
@@ -75,6 +98,14 @@ generate_modulemap = rule(
         "module_name": attr.string(
             doc = "The name of the module.",
         ),
+        "noop": attr.bool(
+            doc = """\
+Designates whether a modulemap should be generated. If `True`, a modulemap is \
+generated. If `False`, a modulemap file is not generated and the returned \
+providers are empty.\
+""",
+            default = False,
```
luispadron, line 107: Should this doc be flipped? if `noop == true` dont we return empty?

cgrindel: Yes. Good catch! I just updated the doc.

PR #728, opened 2023-11-10
chore: extend argument parser example

luispadron:
Extends the `MyExecutable` example to demonstrate a bug when `import RegexBuilder` is used. It seems to be unable to find the dependency label for `RegexBuilder` and then erroneously adds a dep for `"@swiftpkg_swift_argument_parser//:Plugins_GenerateManual"` which fails to build.


luispadron:
@cgrindel I imagine `RegexBuilder` needs to be hooked up like the other Apple APIs?

cgrindel (approved):

PR #729, opened 2023-11-10
feat: Add support RegexBuilder built-in framework

luispadron:
Adds support for built-in frameworks like [RegexBuilder](https://developer.apple.com/documentation/regexbuilder).

### DIFF
--- a/gazelle/internal/swift/is_builtin_module.go
+++ b/gazelle/internal/swift/is_builtin_module.go
@@ -3,6 +3,7 @@ package swift
 import mapset "github.com/deckarep/golang-set/v2"
 
 var otherBuiltinModules = mapset.NewSet[string](
+	"RegexBuilder", // TODO: not included in System/Library/Frameworks for some reason
 	"XCTest",
 )
 


luispadron:
Fixes the issue seen in #728

luispadron:
hmm running `bazel run //tools/generate_builtin_frameworks` removes `RegexBuilder` but updates the file with more frameworks

luispadron:
Added support to the generator for picking up `.swiftmodule`s without `.framework` like the `RegexBuilder` framework

cgrindel (approved):

luispadron:
This is probably having unintended issues elsewhere since it's picking up some other stuff in `/usr/lib/swift`:

```
ld: framework not found os
```

Thoughts @cgrindel

luispadron:
I can also remove the swift module lookup changes and just add RegexBuilder to the other frameworks list, similar to XCTest. I'm not sure why this isn't a framework in System/Frameworks

cgrindel:
> I can also remove the swift module lookup changes and just add RegexBuilder to the other frameworks list, similar to XCTest. I'm not sure why this isn't a framework in System/Frameworks

That sounds reasonable to me.

luispadron:
Updated

cgrindel (approved):

PR #740, opened 2023-11-14
chore: update CI to only configure remote cache auth if secret is set

jpsim:
Because PRs from forks don't have access to secrets from the main repo.

> Due to the dangers inherent to automatic processing of PRs, GitHub’s
> standard `pull_request` workflow trigger by default prevents write
> permissions and secrets access to the target repository.

See https://securitylab.github.com/research/github-actions-preventing-pwn-requests/


cgrindel (approved):

cgrindel:
I am not sure if GH allows `secrets` to be used in the `if` condition. I tried updating your fork, but I did not have permission. So, I copied over the changes to #746.

PR #746, opened 2023-11-15
fix: only configure remote cache if the API key is available

cgrindel:
Copied over a change proposed in #740.


cgrindel:
cc: @jpsim

jpsim (approved):
Thanks for fixing my approach

luispadron (approved):

PR #747, opened 2023-11-15
chore: upgrade to Xcode 15.0.1, Swift 5.9.1

cgrindel:
- Upgrade to Xcode 15.0.1.
- Upgrade to Swift 5.9.1.


.github/actions/set_up_macos/action.yml
```diff
@@ -7,7 +7,7 @@ inputs:
     required: true
   xcode_version:
     description: The version of Xcode to use.
-    default: '14.3.1'
+    default: '15.0.1'
```
luispadron, line 10: Should we run the built-in frameworks update? it was last generated with Xcode 15.0.1, not sure if that gets caught by CI though.

luispadron (approved):

.github/actions/set_up_macos/action.yml
cgrindel: Then, we should be OK. Perhaps we need to tie the generation of the frameworks into CI or just generate as part of the build. 🤔

cgrindel: I added #748 to track.

jpsim (approved):
I'm surprised you didn't need to set the minimum macos version / host version since I was running into issues without that locally.

luispadron:
> I'm surprised you didn't need to set the minimum macos version / host version since I was running into issues without that locally.

I too have been running into issues with that locally and needed to use `--macos_minimum_os=13.0` should we set that in the .bazelrc?

jpsim:
> I too have been running into issues with that locally and needed to use `--macos_minimum_os=13.0` should we set that in the .bazelrc?

I found that 10.15 worked, although CI was still failing but I think that was unrelated:

```
build --macos_minimum_os=10.15 --host_macos_minimum_os=10.15
```

cgrindel:
So, I am a little confused. What changed that requires this flag now?

luispadron:
I believe thats because Xcode 15 just doesn't work on macOS less than 13.0

jpsim:
```
/private/var/tmp/_bazel_runner/3d3cdb00e25c4c831fb244d7b1b9e8aa/execroot/_main/_tmp/22682c64f40ead84b8346b69807cd242/_bazel_runner/68b9399662195c8ac4eff7b8e1251519/external/bazel_tools/tools/build_defs/repo/http.bzl:382:31: in <toplevel>
ERROR: Skipping '//...': error loading package under directory '': error loading package '.build/checkouts/swift-syntax': Label '//utils/bazel:swift_syntax_library.bzl' is invalid because 'utils/bazel' is not a package; perhaps you meant to put the colon here: '//:utils/bazel/swift_syntax_library.bzl'?
ERROR: error loading package under directory '': error loading package '.build/checkouts/swift-syntax': Label '//utils/bazel:swift_syntax_library.bzl' is invalid because 'utils/bazel' is not a package; perhaps you meant to put the colon here: '//:utils/bazel/swift_syntax_library.bzl'?
```

This one is easily solved by adding `.build` to `examples/tca_example/.bazelignore` like I'm doing in https://github.com/cgrindel/rules_swift_package_manager/pull/755

cgrindel:
@jpsim I am perplexed why some of the examples need this and some do not. I added #758 to track this and possibly mitigate having the `.build` stay in the directory. I know @AttilaTheFun would prefer not to have it live in the workspace.

AttilaTheFun:
@jpsim @cgrindel yeah I'm working on a prototype of how we could avoid generating the .build directory in the repository, and avoid checking the swift-deps-index.json into the repository as well. I currently have .build in my .bazelignore and my .gitignore.

cgrindel:
@jpsim Can you remind me what the original purpose of this upgrade is? I feel like Sisyphus trying to get this PR to merge.🙂🪨🗻

jpsim:
> Can you remind me what the original purpose of this upgrade is?

My original motivation was to be able to exercise macros in examples on CI. This PR will do that, but requires CI to have Xcode 15 installed: https://github.com/cgrindel/rules_swift_package_manager/pull/755

Beyond that though, it is generally a good idea to make sure that rules_swift_package_manager is well tested on the latest stable version of Swift & Xcode.

> I feel like Sisyphus trying to get this PR to merge.

I felt that way too in my initial PR, which is why I was happy to let you take the reins 😁

One potential way forward is to set the Xcode version per CI job, where most could default to Xcode 15 but `shake_ios` and `stripe` could use Xcode 14 with a TODO to investigate further why they pass locally but fail on CI.

cgrindel:
> One potential way forward is to set the Xcode version per CI job, where most could default to Xcode 15 but shake_ios and stripe could use Xcode 14 with a TODO to investigate further why they pass locally but fail on CI.

We generate the CI workflow. So, we could have the Xcode version be one of the parameters. 🤔

Issue #742, opened 2023-11-14
Feature request: Better support for built-in framework determination


luispadron:
There are two sub issues I'd like made better:

## Overview

- [x] We are unable to correctly determine when something is a built-in framework.
  - https://github.com/cgrindel/rules_swift_package_manager/pull/749
  - https://github.com/cgrindel/rules_swift_package_manager/pull/750

- [x] We are unable to differentiate between a local framework and a built-in framework of the same name.
  - https://github.com/cgrindel/rules_swift_package_manager/pull/756

## Details 

1. I have several frameworks which import things like `Darwin`, since `Darwin` is not included as a [built-in framework here](https://github.com/cgrindel/rules_swift_package_manager/blob/main/gazelle/internal/swift/builtin_modules.go). This causes a bug similar to the one seen in #728 where the plugin would add a random dep to the target. We need to generate this list so that it's at least more complete.

2. I have a framework named `System`, since `System` is defined as a built-in framework the gazelle generation does not work as expected. It fails to add `System` as a dep of things that use it as it believes its a built-in framework. We need better support for differentiating between frameworks and built-in frameworks with the same name.

cgrindel:
Do you have thoughts on how to mitigate these issues?

1. We can/should add `Darwin` to the [list of special built-ins](https://github.com/cgrindel/rules_swift_package_manager/blob/6023ab6055d225c64cff50f73da87dd78d3491d3/gazelle/internal/swift/is_builtin_module.go#L5-L8).
2. We could add a directive that allows a client to provide their own mapping for select modules.

luispadron:
Yeah so my current thoughts are:

For 1, we could look at updating the list to include swift modules like I had in my PR before undoing that change. It had some other issues but general think it ended up picking a larger set of actual built-in frameworks (im not sure on why the frameworks directory doesnt include these).

For 2, I was hoping this wouldn't have to be configured by users, using `System` as a module name for example should just work. I think this relates a lot to the [discussion you opened already](https://github.com/cgrindel/rules_swift_package_manager/discussions/743). I imagine it's going to be tough to figure this out without directly parsing `Package.swift`

cgrindel:
On second thought, perhaps we could solve issue 2 by introducing a priority. Any modules discovered in a client's workspace should take precedence over any of the others. We could tier it even further, if necessary. 🤔

luispadron:
[...]

PR #735, opened 2023-11-13
feat: add support for Swift macros

jpsim:
* Add support for packages using Swift macros (fixes #700)
* Update `swift-composable-architecture` to 1.4.0 in the TCA example

PR #750, opened 2023-11-15
feat: add more built in frameworks

luispadron:
This resolves the first part of issue #742.

It adds support for finding built-in "frameworks" that are not included as `.framework`s. In addition to the existing framework searching, we now also search in the `usr/lib/swift` directory of the target SDK.

This allows including more builtin frameworks, such as `os` and `Darwin`.

Support is added in a way where the frameworks and modules under `/usr/lib/swift` are split. This is important for supporting `obc_library` and `frameworks` attr. The swift modules are not valid objective-c frameworks.

NOTE: this bumps the SDK version to 14.0 since I generated this file using Xcode 15.0.0. The deletions seem reasonable as they have been deprecated frameworks for a bit.

Depends on: #756


tools/generate_builtin_frameworks/generate_builtin_frameworks.sh
```diff
@@ -36,11 +36,21 @@ sdk_dirs() {
...
+  frameworks=$(find "${frameworks_dir}" -name "*.framework" -depth 1 -not -name "_*" -exec basename -s .framework {} \;)
+
+  # Find all `.swiftmodule` directories within the expected SDK library path.
+  # This includes things like `os` and `Darwin` which do not have `.framework` equivalents.
+  local swift_modules
+  swift_modules=$(find "${lib_swift_dir}" -name "*.swiftmodule" -depth 1 -not -name "_*" -exec basename -s .swiftmodule {} \;)
+
+  echo -e "${frameworks}\n${swift_modules}"
 }
-export -f list_frameworks_for_sdk
```
luispadron, line 43: not sure what this was doing but seemed unused, couldnt find references to it.

tools/generate_builtin_frameworks/generate_builtin_frameworks.sh
```diff
@@ -36,11 +36,21 @@ sdk_dirs() {
 
 list_frameworks_for_sdk() {
```
luispadron, line 37: One other change is, we no longer sort here since it all gets sorted at the end in `list_frameworks_for_platform` anyway.

tools/generate_builtin_frameworks/generate_builtin_frameworks.sh
```diff
@@ -56,7 +66,7 @@ list_frameworks_for_platform() {
     frameworks+="$(list_frameworks_for_sdk "${sdk_path}")"
   done
 
-  echo "${frameworks}" | sort -u
+  echo "${frameworks}" | sort --unique
```
luispadron, line 69: I found `--unique` to be more clear

tools/generate_builtin_frameworks/generate_builtin_frameworks.sh
luispadron, line 1: This file was also auto-formatted by my editor but it seems like it cleaned up some white spaces.

cgrindel (approved):

luispadron:
@cgrindel both `simple_test` and `objc_code_test` seem to be failing in CI but when testing locally they are passing. I did have to pass `--macos_minimum_os=13.0` though to get them to pass is there something I'm missing here?

I cant tell why they'd be failing in CI

cgrindel:
I am merging #747 now with the flags. Should we try rebasing once that and #735 are merged?

luispadron:
@cgrindel Turns out i was able to repro the CI issue and it was legitimate. Adding all swift modules into one builtin frameworks list breaks for Objective-C targets as these are not importable there.

Ive updated the PR to split this into frameworks and swift modules, the code then handles it as expected (only allowing swift modules to be considered built in frameworks for Swift targets)

gazelle/internal/swift/is_builtin_module.go
```diff
@@ -2,30 +2,48 @@ package swift
...
-var otherBuiltinModules = mapset.NewSet[string](
-	"RegexBuilder", // TODO: not included in System/Library/Frameworks for some reason
+var builtinTestFrameworks = mapset.NewSet[string](
 	"XCTest",
 )
 
 // The list of frameworks is found in builtin_modules.go.
-var allBuiltInModuleSets = []mapset.Set[string]{
-	otherBuiltinModules,
+var allBuiltInFramewrokSets = []mapset.Set[string]{
```
cgrindel, line 10: Can you fix the spelling? I think that it should be `allBuiltInFrameworkSets`.

gazelle/internal/swift/is_builtin_module_test.go
```diff
@@ -21,7 +21,22 @@ func TestIsBuiltInModule(t *testing.T) {
...
+
+func TestIsBuiltInSwiftModule(t *testing.T) {
+	tests := []struct {
+		msg  string
+		name string
+		exp  bool
+	}{
+		{msg: "os", name: "os", exp: true},
+		{msg: "does not exist", name: "DoesNotExist", exp: false},
+	}
```
cgrindel, line 37: Could we add one `true` test case?

cgrindel (commented):
Just a couple of comments. Also, it looks like a CI test is failing.

gazelle/internal/swift/is_builtin_module.go
luispadron: Oops yes

gazelle/internal/swift/is_builtin_module_test.go
luispadron: "os" is the true case but let me know if you mean another one

luispadron: That's the best time to do them though! I'll get this rebased and fixed soon

PR #756, opened 2023-11-17
fix: correctly differentiate built-in and local modules with same name

luispadron:
Fixes part of #742.

We previously early-exited if a module was determined to be built-in, we now at least check if there is a local package of that same name


gazelle/resolve.go
```diff
@@ -54,15 +54,14 @@ func (l *swiftLang) Resolve(
...
-			continue
-		}
 		findResults := ix.FindRulesByImportWithConfig(
 			c, resolve.ImportSpec{Lang: swiftLangName, Imp: imp}, swiftLangName)
 		if len(findResults) > 0 {
 			addToDeps(findResults[0].Label)
 			rr.AddLocal(imp, findResults)
+		} else if swift.IsBuiltInModule(imp) {
+			rr.AddBuiltin(imp)
+			continue
```
luispadron, line 64: Im thinking we should move this out after external modules actually. Otherwise we might drop them if they are external but have the same name as a built in module

luispadron: I moved to after external product resolution

gazelle/resolve.go
```diff
@@ -90,7 +89,7 @@ func (l *swiftLang) Resolve(
 	if len(unresolved) > 0 {
 		rr.AddUnresolved(unresolved...)
 		log.Printf("Unable to find dependency labels for %v",
-			strings.Join(resResult.Unresolved, ", "))
+			strings.Join(unresolved, ", "))
```
luispadron, line 92: This was displaying the wrong list because `resResult.Unresolved` is not modified, only `unresolved` is added to if we cant find in `http_archive` and now in built-ins.

bzlmod/workspace/Sources/MyExecutable/BUILD.bazel
```diff
@@ -8,6 +8,7 @@ swift_binary(
     visibility = ["//visibility:public"],
     deps = [
         "//Sources/MyLibrary",
+        "//Sources/System",
```
luispadron, line 11: Before this fix, this dep would never be added

cgrindel (approved):

PR #760, opened 2023-11-20
chore: update CI to use Xcode 15 and Swift 5.9

jpsim:
Except for the `shake_ios` and `stripe` examples, which fail to complete their tests on Xcode 15 for some reason.

Add TODOs to investigate those.


.github/actions/set_up_macos/action.yml
```diff
@@ -5,20 +5,28 @@ inputs:
...
   using: composite
   steps:
     - uses: maxim-lobanov/setup-xcode@v1
       with:
-        xcode-version: ${{ inputs.xcode_version }}
+        xcode-version: '15.0.1'
+    - name: Override Xcode Version
+      shell: bash
+      run: |
+        # TODO: Investigate why these examples fail on CI with Xcode 15
```
cgrindel, line 21:
```suggestion
        # GH762: Investigate why these examples fail on CI with Xcode 15
```
I try to keep TODO comments out of the checked-in code. Instead, I use `GHXXX` comments where the `XXX` is the issue number to track the follow-up work.

cgrindel (approved):
I added #762 to track the follow-up work.

Issue #790, opened 2023-12-07
DataDog 2.5.0 fails to build


jpsim:
Released here: https://github.com/DataDog/dd-sdk-ios/releases/tag/2.5.0

```
external/rules_swift_package_manager~override~swift_deps~swiftpkg_dd_sdk_ios/DatadogTrace/Sources/TraceConfiguration.swift:15:24: error: enum 'URLSessionInstrumentation' does not exist in module 'DatadogInternal'
@_exported import enum DatadogInternal.URLSessionInstrumentation
                       ^               ~~~~~~~~~~~~~~~~~~~~~~~~~
external/rules_swift_package_manager~override~swift_deps~swiftpkg_dd_sdk_ios/DatadogTrace/Sources/Integrations/TracingURLSessionHandler.swift:56:84: error: extra argument 'tracestate' in call
                writer = W3CHTTPHeadersWriter(sampler: tracingSampler, tracestate: [:])
                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~
external/rules_swift_package_manager~override~swift_deps~swiftpkg_dd_sdk_ios/DatadogTrace/Sources/TraceConfiguration.swift:15:24: error: enum 'URLSessionInstrumentation' does not exist in module 'DatadogInternal'
@_exported import enum DatadogInternal.URLSessionInstrumentation
                       ^               ~~~~~~~~~~~~~~~~~~~~~~~~~
```

[2.4.0](https://github.com/DataDog/dd-sdk-ios/releases/tag/2.4.0) builds fine.

cgrindel:
Created #797 to try and reproduce the issue. I added DataDog 2.5.0 to the `resources_example`. The compilation works. Perhaps there is another module that is being referenced in the client's workspace. I have asked @jpsim to check #797 locally. Also, it might be useful to get a list of the imports that are being used from the DataDog package.

jpsim:
I’m not seeing this issue any more. Not sure what it was originally. Sorry for never providing more info.

PR #797, opened 2023-12-09
chore: add DataDog 2.5.0 to resources_example

cgrindel:
Related to #790.


cgrindel:
@jpsim This works for me locally. Can you pull this down and see if it builds for you?

PR #834, opened 2023-12-30
feat: Implement support for new swift_proto_library rule

AttilaTheFun:
This PR implements support in the gazelle plugin for the new swift proto library implementation created in rules_swift:
https://github.com/bazelbuild/rules_swift/pull/1140

There is more context on why the rewrite was necessary and what opportunities the new implementation affords us on that PR.

The new implementation allows us to address the two most significant issues with the current proto support:
- The old implementation used the SwiftProtobuf / GRPC libraries from rules_swift, making them mutually exclusive with SPM-provided versions in the same binary.
- As a workaround to the previous issue, we could generate the targets but could not resolve the GRPC runtime dependencies because they came from rules_swift, so you would see warnings about unknown GRPC, SwiftProtobuf and SwiftNIO targets.

The new implementation fixes these issues by allowing us to declare our own swift_proto_compiler target which uses the SwiftProtobuf and GRPC targets provided through Swift Package Manager / rules_swift_package_manager.

E.g.
```
swift_proto_compiler(
    name = "swift_proto",
    plugin = "@swiftpkg_swift_protobuf//:Sources_protoc-gen-swift",
    plugin_name = "swift",
    plugin_options = BASE_PLUGIN_OPTIONS,
    protoc = "@com_google_protobuf//:protoc",
    suffixes = [".pb.swift"],
    deps = [
        "@swiftpkg_swift_protobuf//:Sources_SwiftProtobuf",
    ],
)
```

These can then be passed into the swift_proto_library targets like so:
```
swift_proto_library(
    name = "echo_service_messages_swift_proto",
    compilers = ["//compilers:swift_proto"],
    module_name = "EchoServiceMessages",
    protos = [":echo_service_messages_proto"],
)
```

This is how the go_proto_library rule works as well.

I updated the gazelle plugin to support generating the new rules, with a directive that allows you to pass these custom compiler targets. I also updated the GRPC example to leverage all of these capabilities.

Now the gazelle plugin can properly resolve swift imports of the proto libraries by their module name, 
E.g. "import EchoServiceMessages", and it can also resolve dependencies between proto libraries mapped to swift proto libraries, so if the code-generated swift imports the same target, it'll receive the same dependency.

NOTE: Due to a bug in rules_swift_package_manager, the GRPC dependencies generated by this repository from the Package.swift are failing to compile on Ubuntu 22.04 in CI.
https://github.com/cgrindel/rules_swift_package_manager/issues/1012

Until this issue is fixed, we need to continue using the GRPC dependencies from rules_swift, but I have split out a follow-up PR to use the SPM version which I can land after the bug is fixed.
https://github.com/cgrindel/rules_swift_package_manager/pull/1019


examples/grpc_example/BUILD.bazel
```diff
@@ -13,10 +13,15 @@ tidy(
...
+# gazelle:proto package
 # gazelle:swift_module_naming_convention pascal_case
 # gazelle:swift_generate_proto_libraries true
-# gazelle:swift_generate_grpc_libraries_with_flavors client,client_stubs,server
+# gazelle:swift_proto_compiler swift_proto=//compilers:swift_proto
+# gazelle:swift_proto_compiler swift_client_proto=//compilers:swift_client_proto
+# gazelle:swift_proto_compiler swift_server_proto=//compilers:swift_server_proto
+# gazelle:swift_generate_grpc_libraries_with_flavors swift_client_proto,swift_server_proto
+# gazelle:exclude .build
+# gazelle:exclude bazel-*
```
cgrindel, line 24: I thought Gazelle was smart enough not to traverse `bazel-*` directories. I have never needed to add this to other projects. Can you explain what prompted this?

cgrindel:
While we wait for the `rules_swift` PR to merge, I am going to convert this to a draft. This helps me keep track of PRs that need immediate attention.

examples/grpc_example/BUILD.bazel
AttilaTheFun: Ah you're right -- I personally set `build --symlink_prefix=.bazel/` in my bazelrc so all of the bazel-* folders are nested under ".bazel" and hidden from finder. Gazelle doesn't handle this correctly, so I have to exclude the .bazel directory. I forgot it already excluded the normal bazel- paths.

AttilaTheFun:
@cgrindel all of my changes have merged into rules_swift now and are tagged in version 1.18.0. We will also be deleting the now-deprecated rules in version 2.0. As such, I have rebased and re-opened this PR now.

gazelle/resolve.go
```diff
@@ -55,8 +97,19 @@ func (l *swiftLang) Resolve(
 	// Try to resolve to targets in this project.
 	externalModules := make([]string, 0, len(swiftImports))
 	for _, imp := range swiftImports {
+		importSpec := resolve.ImportSpec{Lang: swiftLangName, Imp: imp}
+		if l, ok := resolve.FindRuleWithOverride(c, importSpec, "swift"); ok {
+			addToDeps(l)
+			rr.AddLocal(imp, []resolve.FindResult{
+				{
+					Label:  l,
+					Embeds: nil, // TODO: This might be broken -- not sure what to put here.
```
cgrindel, line 106: Is this really a TODO? Can you explain the context?

cgrindel (commented):
First, thank you. I know that this has been a haul. I appreciate the effort. 

Overall, LGTM. I have just one question about a comment in the PR.

gazelle/resolve.go
AttilaTheFun: @cgrindel this was left from when I was initially trying to implement the feature and I wasn't sure what purpose the embeds served and if they were necessary. After some investigation, I found that the "embeds" are kind of a Go-specific concept in Gazelle. If target A embeds target B, only A will be indexed and can be resolved as a dependency by other targets.

There is a method called Embeds on the Resolver interface in gazelle which supplies these, but the implementation for the proto language just returns nil from this method.

```
	// Embeds returns a list of labels of rules that the given rule embeds. If
	// a rule is embedded by another importable rule of the same language, only
	// the embedding rule will be indexed. The embedding rule will inherit
	// the imports of the embedded rule.
	Embeds(r *rule.Rule, from label.Label) []label.Label
```

I don't actually see an implementation of this method at all in rules_swift_package_manager so I'm unsure how it's able to implement the Resolver interface (which is embedded in the Language interface)?

https://github.com/bazelbuild/bazel-gazelle/blob/f5a5c5dc90b4d6a39f0f47d770eba27c6c1c1877/resolve/index.go#L47

In any case, it seems fine to pass nil here, as that's what the proto language does. I've removed the comment here and pushed the commit.

cgrindel (approved):

gazelle/resolve.go
cgrindel: We [start from language.BaseLang](https://github.com/cgrindel/rules_swift_package_manager/blob/4080822c6dc6aed07be6ba8459d9bcf222dcb51e/gazelle/lang.go#L13-L15). This provides reasonable defaults for Gazelle features that are not required for a plugin.

AttilaTheFun: Ah I missed that -- yeah then it's getting the default Embeds implementation which just returns nil:

```
func (b *BaseLang) Embeds(r *rule.Rule, from label.Label) []label.Label {
	return nil
}
```

luispadron (approved):

AttilaTheFun:
Thanks @cgrindel and @luispadron ! Glad to see this finally land.
Once this issue is fixed: https://github.com/cgrindel/rules_swift_package_manager/issues/1012
I'll open this PR: https://github.com/cgrindel/rules_swift_package_manager/pull/1019
Which uses the custom proto compiler feature of the new rules to load the GRPC and SwiftProtobuf dependencies through SPM instead.

PR #875, opened 2024-01-24
fix: handling of `.xcdatamodeld` resources

brentleyjones:
Similar to asset catalogs, we want to include all of the files under `.xcdatamodeld` directories.


swiftpkg/internal/resource_files.bzl
```diff
@@ -6,23 +6,24 @@ load("@bazel_skylib//lib:sets.bzl", "sets")
 # For a list of the auto-discovered resource types, see
 # https://github.com/apple/swift-package-manager/blob/main/Sources/PackageLoading/TargetSourcesBuilder.swift#L634-L677
 _XIB_EXTS = ["nib", "xib", "storyboard"]
-_ASSET_CATALOG_EXTS = ["xcassets"]
```
cgrindel, line 9: Why was this removed?

cgrindel (commented):
Is there a Swift package in the wild that uses `.xcdatamodeld`? If so, perhaps we can add it to the `resources_example`.

swiftpkg/internal/resource_files.bzl
brentleyjones: It's handled by `_is_under_asset_catalog_dir` for actual files under it, and we don't want the directories _themselves_ to ever be valid resources.

brentleyjones:
The example I encountered is https://github.com/mastodon/mastodon-ios/tree/develop/MastodonSDK/Sources/CoreDataStack/CoreData.xcdatamodeld, but I think it's a local-only package.

cgrindel (approved):

PR #982, opened 2024-03-26
feat: Support netrc authentication

jflan-dd:
Support authentication using `.netrc` file when downloading artifacts.

I copied the implementation from here: https://github.com/bazelbuild/bazel/pull/20915/files


cgrindel (approved):
Thank you!

PR #983, opened 2024-03-26
fix: follow symlinks when searching for xcframework directories

jflan-dd:
One of the packages we depend on puts its XCFramework behind a symlink for some reason. Without this change rspm doesn't follow the symlink to find the contents of the actual framework.

mimic behavior of [_list_files_under](https://github.com/cgrindel/rules_swift_package_manager/blob/85a704df7fb25d861ba8e54a584c08e07a7de748/swiftpkg/internal/repository_files.bzl#L44-L45)


cgrindel:
Curious. What was the situation that instigated this change?

jflan-dd:
@cgrindel One of our third party dependencies is laid out with a symlink at the root to an XCFramework that's in its own directory. The `Package.swift` file points to the symlink for the `binaryTarget`'s path.

Similar to this: https://github.com/jflan-dd/symlink_framework

Without this change [this code](https://github.com/cgrindel/rules_swift_package_manager/blob/e5659eab9cf135299ba4b58c728ceb6a560c0d43/swiftpkg/internal/artifact_infos.bzl#L172-L177) wasn't following the symlink to find the `.framework` inside.

cgrindel:
What do you think about adding that Swift package to one of the examples? This will ensure the code change works and it continues to work.

jflan-dd:
@cgrindel I'm very new to Bazel and don't have a good understanding of how the tests are setup in this repository. I've added an example with an empty XCFramework behind a symlink, but I don't know how to add a test to make sure it can build or make sure it runs with the other tests.

Could you provide some guidance?

cgrindel:
@jflan-dd I think we want this example to have a local Swift package that includes the symlinked framework and then some code that uses the Swift package in some way. Here is my suggestion:

1. Create a `third_party/symlinked_framework` directory.
2. Move the `Framework` directory and the `EmptyFramework.xcframework` symlink under `third_party/symlinked_framework`. 
3. Move `swift/Package.swift` to `third_party/symlinked_framework` and modify the path in the file to point to the correct place. Also, change the name of the product to be `EmptyFramework`.
4. Add `Package.swift` to the `examples/symlink_example` directory. It should reference the `EmptyFramework` product in the local package. (see the [resources_example](https://github.com/cgrindel/rules_swift_package_manager/blob/85a704df7fb25d861ba8e54a584c08e07a7de748/examples/resources_example/swift/Package.swift#L8)).
5. Add some Swift code that uses the `EmptyFramework`.
6. Use the Gazelle plugin to generate some Bazel build files.

How does that sound?

cgrindel:
One last thing, once you are able to run `do_test` from inside the `symlink_example`, you should 

1. Run `bazel clean` from inside the `symlink_example`. (Remove the `bazel-XXX` symlinks.)
2. Add your example to [this list](https://github.com/cgrindel/rules_swift_package_manager/blob/227502e641e4b35cbb59928f500ad7037ee609cb/examples/example_infos.bzl#L142-L157).
3. From the parent workspace (`rules_swift_package_manager`), execute `bazel test //examples:symlink_example_test_bazel_.bazelversion`. This executes the example in the same way that it is run in CI.

jflan-dd:
@cgrindel Thanks for the guidance!

I've added a test and validated locally that the test fails without the change to follow symlinks.

examples/symlink_example/do_test
```diff
@@ -0,0 +1,18 @@
...
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" > /dev/null && pwd)"
+
+# Use the Bazel binary specified by the integration test. Otherise, fall back 
+# to bazel.
+bazel="${BIT_BAZEL_BINARY:-bazel}"
+
+# GH249: Do not update build files until the Gazelle extension supports the
+# shape of this example.
+# # Generate Swift external deps and update build files
+# bazel run //:tidy
```
cgrindel, line 14: Is this true for this example? In other words, does //:tidy work for this example?

cgrindel (commented):
Overall, LGTM. Just a question about tidy and the example.

examples/symlink_example/do_test
jflan-dd: It seems to work fine. I had copied the `resources_example` as a starting point for this example.

I pushed a commit to run `//:tidy` as part of `do_test`

jflan-dd:
@cgrindel are there any other changes/additions you're looking for?

cgrindel (approved):
LGTM. Thanks!

PR #985, opened 2024-03-27
fix: exclude paths should be relative to the target path

jflan-dd:
Exclude paths should be relative to the target paths. https://developer.apple.com/documentation/packagedescription/target/exclude

I'm not sure if it's better to compose the paths together here, or further upstream somewhere


jflan-dd:
@cgrindel sorry for all of the terse PRs. I was churning them out to unblock myself while trying to get our app building. I'll work on fleshing out some examples today.

jflan-dd:
@cgrindel here's an example repository where `Error.swift` is not correctly excluded from `DummyPackage` because the path is not treated as relative to the target's path: https://github.com/jflan-dd/exclude_example

cgrindel:
Do you think that we can replicate this in one of our examples? Alternatively, we can add a new example. WDT?

PR #1040, opened 2024-04-24
fix: Don't use legacy UTF8 handling when writing files

jflan-dd:
We have a path that includes a unicode character and without this change the build file is being written with a very strangely encoded path.

According to the documentation the default is going to change soon: https://bazel.build/rules/lib/builtins/repository_ctx#file

### DIFF
--- a/swiftpkg/internal/build_files.bzl
+++ b/swiftpkg/internal/build_files.bzl
@@ -65,7 +65,7 @@ def _write(repository_ctx, build_file, path):
         bld_file_path,
         content = scg.to_starlark(build_file),
         executable = False,
-        legacy_utf8 = False
+        legacy_utf8 = False,
     )
 
 build_files = struct(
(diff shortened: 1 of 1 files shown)


swiftpkg/internal/build_files.bzl
```diff
@@ -65,6 +65,7 @@ def _write(repository_ctx, build_file, path):
         bld_file_path,
         content = scg.to_starlark(build_file),
         executable = False,
+        legacy_utf8 = False
```
brentleyjones, line 68:
```suggestion
        legacy_utf8 = False,
```

cgrindel:
If after accepting @brentleyjones suggested fix, the `tidy_and_test_matrix` should pass. If it does not, please run `bazel run //:tidy`.

cgrindel (approved):
I am approving assuming that you will accept @brentleyjones fix and ensure that the tests are green. Thanks!

PR #957, opened 2024-03-10
feat: no longer need to check in `swift_deps_index.json` file and other big updates

cgrindel:
Lots of big changes in this PR:

- The SPM-related rules no longer depend upon a Swift deps index JSON file.
- Remove the `from_file` module extension tag class, as the ruleset no longer needs to read the Swift deps index file.
- Remove the `swift_update_packages` macro as it is no longer required.
- Remove support for `byName` references outside of a Swift package. This was supported in early SPM versions. It was one of the driving factors for having to pre-process the Swift dependencies and generate a Swift deps index file. Removing this support reduces the complexity of the `rules_swift_package_manager` implementation.
- Remove `update_repos` command from the Gazelle plugin.
- Add `swift_deps_info` repository rule to generate a Swift deps index JSON file as part of the build, not in the Gazelle plugin. This index file is used by the Gazelle plugin to resolve Swift module names to their respective Bazel target.
- Introduce `from_package` module extension tag class telling the ruleset where the `Package.swift` and `Package.resolved` files for the workspace live. These are used to identify the Swift package transitive dependencies for the project.
- Remove the `http_archive_ext_deps` example. The example existed to demonstrate and test indexing of Swift targets that were downloaded using `http_archive`. The Gazelle internals no longer provide some of the information needed for this to work properly with bzlmod enabled. Given that SPM has been widely adopted and there are some workarounds, we are removing support for this functionality in the Gazelle plugin when bzlmod is enabled.
- Add output to the example test runner that demarcates different phases for the testing.

Related to #924.

PR #1043, opened 2024-04-25
feat: Swift registry support

### LINKED ISSUE #1016: Swift registry support
It looks like there's currently no support for package dependencies resolved from a [package registry](https://github.com/apple/swift-package-manager/blob/main/Documentation/PackageRegistry/PackageRegistryUsage.md). That is, packages defined with the  `id` syntax:

```swift
dependencies: [
    .package(id: "mona.LinkedList", .upToNextMajor(from: "1.0.0")),
],
```

If Package.swift contains a dependency like this, the `swift_update_pkgs` task fails with this error:
```
gazelle: Identity could not be determined.
```

From a cursory look at the source it seems like registry dependencies aren't part of the data model. I'm not sure what additional changes would be necessary on top of that.

I'm guessing it might be a pain to build this out if one doesn't have access to a registry already, since there are no public Swift registries I am aware of. I'm happy to help test things or provide samples from a project resolved against a registry if it's useful.

watt:
Fixes #1016 

This PR adds support for package registry dependencies, by adding a new build rule type. For a Package.swift dependency like this:

```swift
    dependencies: [
        .package(id: "apple.swift-collections", from: "1.1.0"),
    ]
```

A build rule will be generated like this:

```bazel
    swift_registry_package(
        name = "swiftpkg_apple.swift_collections",
        dependencies_index = "@//:swift_packages_index.json",
        id = "apple.swift-collections",
        version = "1.1.0",
    )
```


examples/registry_example/README.md
```diff
@@ -0,0 +1 @@
+# registry Example
```
luispadron, line 1: Lets provide more details here on what this is meant showcase and any relevant files / code.

examples/registry_example/swift_deps.bzl
```diff
@@ -0,0 +1,8 @@
+load("@rules_swift_package_manager//swiftpkg:defs.bzl", "swift_registry_package")
+
+def swift_dependencies():
+    swift_registry_package(
+        name = "swiftpkg_apple.swift_collections",
+        dependencies_index = "@//:swift_packages_index.json",
+        url = "https://artifactory.global.square/artifactory/api/swift/swift-test/apple/swift-collections/1.1.0.zip",
```
luispadron, line 7: This should be a public URL so we can test in CI

gazelle/internal/spdump/dependency.go
```diff
@@ -167,3 +171,22 @@ func (fs *FileSystem) UnmarshalJSON(b []byte) error {
 	fs.Path = rawFS.Path
 	return nil
 }
+
+type reg struct {
+	Identity string
+}
+
+type Registry struct {
```
luispadron, line 179: nit: could we add a comment describing the purpose of this type like `FileSystem`

gazelle/internal/spreso/pin_store_v2.go
```diff
@@ -2,7 +2,7 @@ package spreso
 
 import (
 	"encoding/json"
-	"fmt"
+	// "fmt"
```
luispadron, line 5:
```suggestion
```

gazelle/internal/spreso/pin_store_v2.go
```diff
@@ -100,9 +100,10 @@ func NewPkgRefFromV2Pin(v2p *V2Pin) (*PackageReference, error) {
...
-	if ps.Revision == "" {
-		return nil, fmt.Errorf("revision cannot be empty %+v", ps)
-	}
+	// revision is empty for registry pins
+	// if ps.Revision == "" {
+	// 	return nil, fmt.Errorf("revision cannot be empty %+v", ps)
+	// }
 	if ps.Version != "" {
 		return NewVersionPinState(ps.Version, ps.Revision), nil
 	}
```
luispadron, line 109: Could we check `ps.Version` first to circumvent this

gazelle/internal/swift/code_dir.go
```diff
@@ -40,3 +40,9 @@ func CodeDirForLocalPackage(pkgDir string, localPkgPath string) string {
 	// Return the local path
 	return filepath.Clean(path)
 }
+
+func CodeDirForRegistryPackage(buildDir string, registryIdentity string, version string) string {
```
luispadron, line 44: nit: add a doc comment describing what this does with an example mapping of registry id to path

gazelle/internal/swift/repo_rule_from_bazel_repo.go
```diff
@@ -25,9 +26,13 @@ func RepoRuleFromBazelRepo(
 	var r *rule.Rule
 	var err error
 	if bzlRepo.Pin != nil {
-		r, err = repoRuleFromPin(bzlRepo.Name, bzlRepo.Pin, patch)
-		if err != nil {
-			return nil, err
+		if bzlRepo.Pin.PkgRef.Kind == spreso.RegistryPkgRefKind {
+			r = repoRuleForRegistryPackage(bzlRepo.Name, bzlRepo.Pin)
```
luispadron, line 30: We should handle any error here as well

luispadron: Potentially this `if` could happen before `bzlRepo.Pin` to reduce nesting too.

swiftpkg/internal/swift_registry_package.bzl
```diff
@@ -0,0 +1,74 @@
...
+    prefix = "{}.{}".format(scope, name)
+
+    headers = { 
+        "Accept": "application/vnd.swift.registry.v1+zip",
+    }
+
+    repository_ctx.download_and_extract(
+        url = url,
+        # requires bazel 7.1
+        headers = headers,
```
luispadron, line 36: This is probably fine for a new feature, we should call this out in any documentation though. I wonder if folks on Bazel < 6 but > 5 could use the credentials helper to work around it?

swiftpkg/internal/swift_registry_package.bzl
```diff
@@ -0,0 +1,74 @@
...
+    directory = str(repository_ctx.path("."))
+    env = repo_rules.get_exec_env(repository_ctx)
+    repo_rules.check_spm_version(repository_ctx, env = env)
+
+    url = repository_ctx.attr.url
+
+    # Works for my case, but the prefix can actually be anything, so this is not sufficient for general use.
+    # See `archive-source` command docs:
+    # https://github.com/apple/swift-evolution/blob/main/proposals/0292-package-registry-service.md#archive-source-subcommand
+    components = url.split("/")
```
luispadron, line 24: What do we need to make this work more generally?

luispadron (commented):
This is a great start thank you!

cgrindel:
Thank you for the contribution! Let me know what you think about [my comment on the issue](https://github.com/cgrindel/rules_swift_package_manager/issues/1016#issuecomment-2079417370).

examples/registry_example/swift_deps.bzl
watt: > This should be a public URL so we can test in CI

This will be tough, because there's no public registry out there. We might be able to create a mock registry in python, and spin that up on CI during tests?

swiftpkg/internal/swift_registry_package.bzl
watt: > This is probably fine for a new feature, we should call this out in any documentation though. I wonder if folks on Bazel < 6 but > 5 could use the credentials helper to work around it?

I'm not familiar with the credentials helper; can it override the `Accept` header?

swiftpkg/internal/swift_registry_package.bzl
watt: > What do we need to make this work more generally?

Something like `stripPrefix` that doesn't require you to know the prefix in advance. It could be something along the lines of:
1. download and extract into a temp directory
2. sanity check there is only one directory in the extracted file location
3. move all the files from that one directory to the `repository_ctx` root (where `download_and_extract` would have put them by default)

Not too difficult to script, but I don't know if there's a way to do it without shelling out to commands.

gazelle/internal/spreso/pin_store_v2.go
```diff
@@ -100,9 +100,10 @@ func NewPkgRefFromV2Pin(v2p *V2Pin) (*PackageReference, error) {
 }
 
 func NewPinStateFromV2PinState(ps *V2PinState) (PinState, error) {
-	if ps.Revision == "" {
-		return nil, fmt.Errorf("revision cannot be empty %+v", ps)
-	}
+	// revision is empty for registry pins
+	// if ps.Revision == "" {
+	// 	return nil, fmt.Errorf("revision cannot be empty %+v", ps)
+	// }
```
cgrindel, line 106: Please remove the commented code.

swiftpkg/internal/swift_registry_package.bzl
```diff
@@ -0,0 +1,88 @@
...
+load("@bazel_tools//tools/build_defs/repo:utils.bzl", "update_attrs")
+load(":pkg_ctxs.bzl", "pkg_ctxs")
+load(":registry_configs.bzl", "registry_configs")
+load(":repo_rules.bzl", "repo_rules")
+load(":repository_files.bzl", "repository_files")
+
+def _remove_bazel_files(repository_ctx, directory):
+    files = ["BUILD.bazel", "BUILD", "WORKSPACE", "WORKSPACE.bazel"]
+    for file in files:
+        repository_files.find_and_delete_files(repository_ctx, directory, file)
```
cgrindel, line 13: Let's move this into `repository_files` and then update `swift_package` and this code to use it.

swiftpkg/internal/registry_configs.bzl
```diff
@@ -0,0 +1,34 @@
+"""Module for resolving registry URLs from scopes."""
+
+def _read(repository_ctx):
+    registry_config_json = repository_ctx.read(
+        Label("@//:.swiftpm/configuration/registries.json"),
```
cgrindel, line 5: We can add an attribute defined as `attr.label(allow_single_file = True)` to `swift_registry_package()`. We can read the value in `_swift_registry_package_impl()` and pass the path to this function.

cgrindel (changes requested):

examples/registry_example/swift_deps.bzl
watt: I just ended up removing this example altogether, since it really only works on my machine.

gazelle/internal/spdump/dependency.go
watt: done!

gazelle/internal/spreso/pin_store_v2.go
watt: Yes! I reordered these and was able to keep the `ps.Revision == ""` guard.

gazelle/internal/swift/code_dir.go
watt: done!

gazelle/internal/swift/repo_rule_from_bazel_repo.go
watt: `repoRuleForRegistryPackage` doesn't return an error. I did flatten this out, but it means having a redundant `bzlRepo.Pin != nil` check.

swiftpkg/internal/swift_registry_package.bzl
watt: I scripted out a more general solution.

gazelle/internal/spreso/pin_store_v2.go
watt: done!

swiftpkg/internal/swift_registry_package.bzl
watt: done!

swiftpkg/internal/registry_configs.bzl
watt: OK, I created a `registry_config` attribute for this. It still defaults to `Label("@//:.swiftpm/configuration/registries.json")`, rather than emitting that to the generated rules, though. Is that OK? I'm worried that letting this be customized is a bit dangerous, because putting any other value in here means it won't be using the same config that `swift package resolve` does.

I think we might want to also pass a `--config-path` arg to `swift` to prevent it from using global config outside the workspace.

swiftpkg/internal/swift_registry_package.bzl
```diff
@@ -0,0 +1,103 @@
...
+        doc = "The package identifier.",
+    ),
+    "version": attr.string(
+        mandatory = True,
+        doc = "The package version.",
+    ),
+    "registry_config": attr.label(
+        default = Label("@//:.swiftpm/configuration/registries.json"),
+        allow_single_file = True,
+        doc = "Path to the project-level Swift registry config JSON file.",
```
cgrindel, line 87: Should we add some verbiage how the registry is generated?

swiftpkg/internal/registry_configs.bzl
cgrindel: Sorry for the delay. I had many family obligations over the past 2 weeks. 

Regarding the label default, that is fine. In another comment, I suggest that we update the doc for the attribute stating where the file comes from / how it is typically generated.

>I think we might want to also pass a --config-path arg to swift to prevent it from using global config outside the workspace.

Where do you want to add that?

watt: > > I think we might want to also pass a --config-path arg to swift to prevent it from using global config outside the workspace.
> 
> Where do you want to add that?

In the `SwiftBin` wrapper that calls `swift package resolve`. By passing a `--config-path` to that command, we can make `swift` only look at the config in this workspace (which is the only one RSPM is looking at). So the command would become something like:
```sh
swift package resolve --config-path .swiftpm/configuration
```

`--config-path` sets the location of the _shared_ config, which is usually `~/.swiftpm/configuration`. The idea here is just to make it identical to the project config so that there's only one config to consider. Setting it to something like `/dev/null` would probably have the same effect (but is maybe more fragile if a new version of swift decides the directly must exist).

swiftpkg/internal/swift_registry_package.bzl
watt: Sure thing. I just pushed a commit that elaborates a bit, let me know if there's more you'd like to see.

swiftpkg/internal/swift_registry_package.bzl
brentleyjones: I don't like requiring Bazel 7.1. I'm also not familiar with the credentials helper. @luispadron do you know an answer to their question?

swiftpkg/internal/registry_configs.bzl
cgrindel: OK. I am fine with having the new parameter passing the `--config-path` value. However, as I may have mentioned in another thread, the `swift_update_packages` macro may disappear before 1.0. I will open a discussion about that shortly.

brentleyjones:
@watt @cgrindel What's needed to land this? Ideally this would be part of 1.0.

cgrindel:
> What's needed to land this? Ideally this would be part of 1.0.

We just need to understand how this PR changes with `swift_update_packages` disappearing from `rules_swift_package_manager`.

watt:
> > What's needed to land this? Ideally this would be part of 1.0.
> 
> We just need to understand how this PR changes with `swift_update_packages` disappearing from `rules_swift_package_manager`.

Is this something I can help with? I'd love to land this as it works today and then iterate towards whatever shape 1.0 has, but I don't have a good sense for what that entails.

cgrindel:
@watt I just merged #957 which removes the `swift_update_packages` and some other big changes. Can you rebase this PR from `main` and raise any issues that you may need help navigating?

cgrindel:
I am sorry that the recent changes thwarted your efforts. Once the 1.0 release happens, it would be great to pick this up again.

PR #1044, opened 2024-04-25
Fix deduplication of -enable-experimental-feature

Dreksh:
When using a Swift package that contains multiple experimental features (i.e. https://github.com/sersoft-gmbh/swift-smtp ), the error that is returned shows treats one of the features as an input file.

This is due to the generated parameters within the `*.swiftmodule-0.params` file that is used to set the build arguments, where `-enable-experimental-feature` was deduplicated. The deduplication happens during `bzl_selects.to_starlark` as it operates on sets, instead of lists.

Changes:
---------
- Introduced `bzl_selects.unwrap_list` to operate on lists instead of sets
- Changed the copts to rely on `unwrap_list`, as it shouldn't be treated as a set.


swiftpkg/internal/swiftpkg_build_files.bzl
```diff
@@ -106,7 +106,7 @@ def _swift_target_build_file(pkg_ctx, target):
     if len(defines) > 0:
         attrs["defines"] = bzl_selects.to_starlark(defines)
     if len(copts) > 0:
-        attrs["copts"] = bzl_selects.to_starlark(copts)
+        attrs["copts"] = bzl_selects.unwrap_list(copts)
```
cgrindel, line 109: Is there any way to detect this in `bzl_selects.to_starlark()`? It is not obvious to me why `copts` would get this special handling.

cgrindel (commented):
Thank you for the contribution! Please see my comment asking whether we can handle this inside `to_starlark()`. Also, could we add https://github.com/sersoft-gmbh/swift-smtp to one of the examples so that we reproduce the issue and confirm that it stays fixed?

swiftpkg/internal/swiftpkg_build_files.bzl
Dreksh: The reason is that we add a string repeatedly:
https://github.com/cgrindel/rules_swift_package_manager/blob/d671d9e8b5f2fc7e5c65eb9bbe072cf62d04d387/swiftpkg/internal/swiftpkg_build_files.bzl#L103

`to_starlark` relies on using sets within its processing, which will deduplicate the string that's added.

I'm not familiar with the repository, so I decided to add a new function to reduce the chances that I'll affect other places that use `to_starlark`.

Possible Alternatives:
----------------------
- Add a parameter like `dedup = True`, and only perform deduplication after the list has been created, so that if `dedup = False` was passed, it will return the list without deduping.
- Move adding `-enable-experimental-feature` after running `to_starlark`, so that it skips the deduplication. This allows us to reuse `to_starlark` as it is now. Experimental features will need to be placed in a different list, so it doesn't affect the other `copts` from unsafe flags.

Dreksh: I tried doing `-enable-experimental-feature` after running `to_starlark`, but didn't realise that it returns a structure instead of a list. So this method won't work, as `to_starlark` will have copts in an internal structure. (I did not dig into whether it'll be easy to append values into that list).

I'll give adding the dedup parameter a try (maybe renamed to `preserve_list`?)

cgrindel:
@Dreksh Were you going to pick this up again? If not, I am inclined to close this. We can always open a new PR when/if you want to continue.

cgrindel:
Alternatively, we could convert it to a Draft PR.

PR #1072, opened 2024-05-14
feat: use suffixed repo name as package name

### LINKED ISSUE #1057: Support for `package` access modifier
[SE-0386](https://github.com/apple/swift-evolution/blob/main/proposals/0386-package-access-modifier.md) introduced the `package` access level. Packages using this require a `-package-name` argument to the compiler, which `rules_swift` rules [support via a `package_name` argument](https://github.com/bazelbuild/rules_swift/blob/master/doc/rules.md).

SE-0386 says this about the package name:

> The build system should make a best effort to ensure that package names are unique. The Swift Package Manager already has a concept of a package identity string for every package. This string is verified to be unique, and it already works as a package name, so SwiftPM will pass it down automatically. Other build systems such as Bazel may need to introduce a new build setting for a package name.

Since RSPM translates rules from SwiftPM, I think using the SwiftPM package name from `Package.swift` directly is the most obvious thing to do here. It should be fairly straightforward to copy this to the `package_name` arg when generating swift rules.

watt:
Resolves #1057.

Following the recommendation from the discussion there, this uses the repo name suffixed with `.rspm` as the package name.


swiftpkg/internal/swiftpkg_build_files.bzl
```diff
@@ -49,6 +49,7 @@ def _swift_target_build_file(pkg_ctx, target):
     attrs = {
         "deps": bzl_selects.to_starlark(deps),
         "module_name": target.c99name,
+        "package_name": "rspm." + target.label.repository_name,
```
brentleyjones, line 52: I would still prefer a suffix please.

cgrindel (commented):
From my quick read, it looks like passing `package_name` will be a noop if the package does not use the `package` access modifier. Is that correct?

swiftpkg/internal/swiftpkg_build_files.bzl
watt: > I would still prefer a suffix please.

changed!

watt:
> From my quick read, it looks like passing `package_name` will be a noop if the package does not use the `package` access modifier. Is that correct?

That's my understanding, yes. I believe SwiftPM does this by default, using the package name in Package.swift.

cgrindel (approved):
LGTM. Thanks!

brentleyjones (approved):

watt:
Seems that downstream compilation workers got confused by the @-prefixed package name and that caused tests to fail in CI. I've tweaked this slightly to trim the leading @.

luispadron (approved):

watt:
I'm having trouble confirming locally, but it seems like the failing tests are running with Xcode 14.3/Swift 5.8, which doesn't support the `-package-name` compiler argument. Is the expectation that `rules_swift_package_manager` should gate on Swift version, or should `rules_swift` be handling that?

brentleyjones:
rules_swift should ignore that attribute when it's not supported. If it doesn't, we should fix it.

swiftpkg/internal/swiftpkg_build_files.bzl
```diff
@@ -49,6 +49,7 @@ def _swift_target_build_file(pkg_ctx, target):
     attrs = {
         "deps": bzl_selects.to_starlark(deps),
         "module_name": target.c99name,
+        "package_name": target.label.repository_name.lstrip("@") + ".rspm",
```
luispadron, line 52: Is it possible to set this to something custom in SPM today? I know it is in Xcode via `PACKAGE_ACCESS_IDENTIFIER`. Just wondering if we would ever run into issues where one package defines `package` visible APIs that are then used by another package.

watt: > Is it possible to set this to something custom in SPM today? I know it is in Xcode via `PACKAGE_ACCESS_IDENTIFIER`. Just wondering if we would ever run into issues where one package defines `package` visible APIs that are then used by another package.

No, not via SwiftPM; it uses the package name from Package.swift.

watt:
> **rules_swift** should ignore that attribute when it's not supported. If it doesn't, we should fix it.

@brentleyjones Should I open an issue on `rules_swift`? Not sure what the next step ought to be here.

luispadron:
I can look at fixing that tomorrow if no one else has time, but rules_swift releases are sort of blocked right now as we get some other stuff is merged so not sure when this repo could pick that change up

luispadron:
Could/should we flag this in this report based on the swift tools version of the Package.swift?

cgrindel:
> it probably makes more sense to gate the setting of `package_name` in this repository and not in rules_swift

SGTM. What is the correct gate? The Swift tools version?

luispadron:
Yeah I imagine that's how it works in SPM but haven't checked

watt:
Just pushed an update that gates `package_name` behind the tools version. The semver parsing is naive but hopefully sufficient for this purpose. If not, please lmk!

PR #1094, opened 2024-06-01
fix: specify all defines as local using `copts`

### LINKED ISSUE #1093: Propagates preprocessor SWIFT_PACKAGE
I have a target that when it depends on an SPM package, the preprocessor definition `SWIFT_PACKAGE` becomes true.

I think that the expected behavior should have a value of false.

cgrindel:
In an attempt to localize defines to specific targets, we specify defines/macros for all target types (e.g., `swift_library`, `objc_library`, and `cc_library`) targets using  `copts`.

Fixes #1093.

### DIFF
--- a/swiftpkg/internal/swiftpkg_build_files.bzl
+++ b/swiftpkg/internal/swiftpkg_build_files.bzl
@@ -71,12 +71,13 @@ def _swift_target_build_file(pkg_ctx, target):
             deps_without_plugins = [dep for dep in deps if dep.value[0] not in plugins]
             attrs["deps"] = bzl_selects.to_starlark(deps_without_plugins)
 
-    defines = [
+    # NOTE: We specify defines using copts so that they stay local to the
+    # target. Specifying them using the defines attribute will propagate them.
+    copts = [
         # SPM directive instructing the code to build as if a Swift package.
         # https://github.com/apple/swift-package-manager/blob/main/Documentation/Usage.md#packaging-legacy-code
-        "SWIFT_PACKAGE",
+        "-DSWIFT_PACKAGE",
     ]
-    copts = []
 
     # GH046: Support plugins.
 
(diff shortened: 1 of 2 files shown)


brentleyjones:
We can emulate `local_defines` with `copts` though. We should do so on both `swift_library` and `objc_library`.

cgrindel:
> We can emulate local_defines with copts though. We should do so on both swift_library and objc_library.

@brentleyjones Can you elaborate on that? Do you mean specifying `-DSWIFT_PACKAGE=1` directly?

brentleyjones:
Yes :+1:. But without `=1` for `swift_library`.

cgrindel:
So, do we want to keep all defines local or just the `SWIFT_PACKAGE`? [We apply other define values, as well.](https://github.com/cgrindel/rules_swift_package_manager/blob/63f615e4882f7d02498c202877f6c671f4cf4da4/swiftpkg/tests/swiftpkg_build_files_tests.bzl#L642-L646) They can be specified as [build settings](https://github.com/apple/swift-package-manager/blob/main/Documentation/PackageDescription.md#methods-10). I am checking the source to see what it does, because the doc is not clear.

cgrindel:
I don't see any code that propagates the define values. So, I will update the code to set all defines as local.

@brentleyjones Is there any background why the `defines` for `swift_library` propagate? Is it just matching the `cc_library` behavior?

brentleyjones:
Yes. It would make more sense to behave that way if you could have a source-less `swift_library`, but at least it's consistent with `cc_library`.

cgrindel:
@brentleyjones @luispadron This PR is ready for review. It now uses `copts` for defines.

brentleyjones (approved):

luispadron (approved):

PR #1102, opened 2024-06-07
fix: improve xcframework archive file link type detection

### LINKED ISSUE #1084: Unable to determine link type for GoogleMaps iOS SDK
Hey everyone!

I'm currently getting the following error while trying to integrate the [Google Maps SDK](https://github.com/googlemaps/ios-maps-sdk):
```
Error in fail: Unexpected link type for target. target: GoogleMaps, link_type: unknown, expected: static, dynamic
```

It seems to be because the result of `file --brief` on the 2 `.a` in `GoogleMaps.xcframework` results in the outputs:
```
ios-arm64/GoogleMaps.a: Mach-O 64-bit object arm64
```
```
Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit object x86_64] [arm64]
ios-arm64_x86_64-simulator/GoogleMaps.a (for architecture x86_64):	Mach-O 64-bit object x86_64
ios-arm64_x86_64-simulator/GoogleMaps.a (for architecture arm64):	Mach-O 64-bit object arm64
```

The [code handling the result](https://github.com/cgrindel/rules_swift_package_manager/blob/main/swiftpkg/internal/artifact_infos.bzl#L129-L140) is expecting either `ar archive` or `dynamic`, which is not the case and so defaults to `unknown`.

I'm not sure what the fix could be - [this post](https://forums.developer.apple.com/forums/thread/736719) suggests using `otool -l` and checking for a `LC_ID_DYLIB` for a dynamic library, maybe that can be used as a more robust check for dynamic libraries, and then maybe if a library doesn't have this it could be considered `static`?

I'd be glad to make that change if that's an acceptable solution

stephanecopin:
- If RSPM cannot infer the link type from `file --brief`, use `otool -l` to see if the library is `dynamic`, and if not default to `static`.
- Add a subset of the [Google Maps Demos](https://github.com/googlemaps-samples/maps-sdk-for-ios-samples/tree/main/GoogleMaps-Swift) to the project.
	- The example needed to be patched, as there's an error with `@import GoogleMaps;` in the headers for the frameworks otherwise.
	FYI, the error is:
    ```
	external/rules_swift_package_manager~~swift_deps~swiftpkg_ios_maps_sdk/Core/GMSEmpty.m:17:9: fatal error: module 'GoogleMaps' not found
	@import GoogleMaps;
	 ~~~~~~~^~~~~~~~~~
	1 error generated.
    ```
	I quickly looked into it but couldn't find an easy fix, so patching seemed like the easiest solution.

cc @cgrindel - would love feedback on this! I didn't do any processing on the `otool -l` command as I didn't think it to be necessary.

Closes #1084


cgrindel (approved):
LGTM. Thank you !

cgrindel:
@stephanecopin I ran `bazel run //:tidy` on this branch and updated it. This got the new integration test running. However, there appears to be [some absolute paths from your machine somewhere in the code](https://github.com/cgrindel/rules_swift_package_manager/actions/runs/9429180619/job/25975292192?pr=1102).

stephanecopin:
Ah my bad, that was leftover code from when I was investigating the `@import` issue 😅 I've removed it, should all good now

cgrindel:
It looks like you do not have any test targets. Can you add one that launches the iOS app or at least a build_test that ensures that the iOS app builds?

stephanecopin:
@cgrindel ah right, I've added a `build_test` as the example will not launch without a valid Google Maps API key.

cgrindel:
It looks like you forgot to add `load("@bazel_skylib//rules:build_test.bzl", "build_test")` to the file with the `build_test`.

stephanecopin:
Ah my bad! That's what you get for doing that late 😅 I've updated & verified it passes locally, so hopefully should be all good now

PR #1105, opened 2024-06-11
fix: parsing modulemaps with comments

luispadron:
Fixes the tokenizer and modulemap parser to support C99 formatted comments. This is probably not amazing but we'll be able to move off this soon: https://github.com/bazelbuild/rules_swift/pull/1212


cgrindel (approved):
LGTM. Thanks!

PR #1107, opened 2024-06-14
fix: Use `swift_interop_hint` instead of `swift_module` tag

brentleyjones:
Needed for rules_swift 2.0+. Only works with rules_swift 2.0+.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -23,12 +23,12 @@ bazel_dep(name = "platforms", version = "0.0.6")
 # rules_swift_package_manager _must_ be a runtime dependency of rules_swift_package_manager.
 bazel_dep(
     name = "rules_swift",
-    version = "1.16.0",
+    version = "2.1.1",
     repo_name = "build_bazel_rules_swift",
 )
 bazel_dep(
     name = "rules_apple",
-    version = "3.1.1",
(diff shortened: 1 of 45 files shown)


brentleyjones:
I can't think of a good way to determine if we are depending on rules_swift 2.0 or not, so right now this is unconditional. Ideally we would only make this change if transitively depending on rules_swift 2.0.

luispadron (approved):

cgrindel (approved):

brentleyjones:
Needs https://github.com/cgrindel/rules_swift_package_manager/pull/1156.

cgrindel (approved):
Thanks!

cgrindel:
I see this in the firbase example failure:
```log
ERROR: rules_apple@3.6.0 depends on rules_swift@1.18.0 with compatibility level 1, but <root> depends on rules_swift@2.1.0 with compatibility level 2 which is different
```

Is there a corresponding rules_apple upgrade that needs to be applied?

brentleyjones:
It's a bad error message (https://github.com/bazelbuild/bazel/issues/22972), it was actually **rules_xcodeproj** that needed to be updated.

luispadron (approved):

PR #1117, opened 2024-06-26
fix: fix `byName` resolution

brentleyjones:
In the case of `accessibilitysnapshot`, it depends on `SnapshotTesting` by name. `SnapshotTesting` has an identity of `swift-snapshot-testing` (based on the git url, but it could be different for mirrors/forks), but the name is `snapshottesting`.

### DIFF
--- a/swiftpkg/internal/pkginfo_target_deps.bzl
+++ b/swiftpkg/internal/pkginfo_target_deps.bzl
@@ -81,7 +81,7 @@ def _resolve_by_name(pkg_ctx, name):
 
     ext_dep = lists.find(
         pkg_ctx.pkg_info.dependencies,
-        lambda d: d.identity == normalized_name,
+        lambda d: d.name == normalized_name,
     )
     if ext_dep != None:
         return [bazel_labels.new(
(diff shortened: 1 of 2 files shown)


swiftpkg/internal/pkginfo_target_deps.bzl
```diff
@@ -81,15 +81,15 @@ def _resolve_by_name(pkg_ctx, name):
...
+        lambda d: d.name == normalized_name,
     )
     if ext_dep != None:
         return [bazel_labels.new(
             name,
             repository_name = bazel_repo_names.from_identity(ext_dep.identity),
             package = "",
         )]
-    fail("Unable to resolve byName reference {name} in {repo_name}.".format(
+    fail("Unable to resolve byName reference {name} in {repo_name}. {deps}".format(
```
cgrindel, line 92: Do we need to add the `deps` to the format arguments?

brentleyjones: No, this was debug only, sorry, will remove.

brentleyjones:
That's exactly what this change is doing, matching on package name. It was previously trying to match on package identity, which is based on the URL of the package.

swiftpkg/internal/pkginfo_target_deps.bzl
brentleyjones, line 92:
```suggestion
    fail("Unable to resolve byName reference {name} in {repo_name}.".format(
```

cgrindel (approved):

luispadron (approved):

brentleyjones:
@cgrindel after this is merged can we get a 0.34.1 release?

PR #1173, opened 2024-07-16
fix: generate a `select` statement for each condition

### LINKED ISSUE #1111: Ambiguous `copt` in release mode
I was updating my dependencies and ran into a build failure **in release mode (e.g. `--compilation_mode=opt`)**

```
ERROR: /private/var/tmp/_bazel_runner/859a0f9921c91b89cc2a6b233b4f521c/external/rules_swift_package_manager~~swift_deps~swiftpkg_swift_system/BUILD.bazel:23:14: Illegal ambiguous match on configurable attribute "copts" in @@rules_swift_package_manager~~swift_deps~swiftpkg_swift_system//:SystemPackage.rspm:
```

roya1v:
This is should fix #1111 . I had the same issue where the `select` in `cops` would have multiple matching conditions, which I think isn't something wrong. So in here I changed the code so it generates separate selects for each option.

I'm not completely happy with the implementation and I still have to fix the tests probably, but please tell me if you think I'm going in a completely wrong direction


cgrindel:
Instead of creating a separate `to_multiple_starlark()` function, can we integrate this logic into the existing `to_starlark()`?

roya1v:
Hey @cgrindel ! Sorry for the long wait, I've combined the two functions like you suggested, let me know what you think

cgrindel (commented):
Overall, looks good. It appears that some unit tests need to be updated.

@luispadron Does this solution solve the original issue that you filed?

swiftpkg/internal/bzl_selects.bzl
```diff
@@ -181,18 +182,33 @@ def _to_starlark(values, kind_handlers = {}):
...
+            if len(expr_members) > 0:
+                expr_members.append(scg.new_op("+"))
+            new_select_dict = {
+                k: sets.to_list(select_dict[k])
+                for k in sorted_keys
+            }
+            if kind_handler.default != None:
+                new_select_dict[_bazel_select_default_condition] = kind_handler.default
+            select_fn = scg.new_fn_call("select", new_select_dict)
+            expr_members.append(select_fn)
```
luispadron, line 211: Maybe we should move this to an internal function so we can reuse the logic in both branches at seems similar across the if/else. Then the only difference would be the for loop

luispadron (commented):
Thanks for working on this, sorry it's taken so long to review!

I'm out of office this week but I can double check if this solves our issue soon, if I forget please remind me!

It would also be nice to either figure out how to test this edge case in a fixture or to also build our current examples in release mode to see if that captures this case already.

swiftpkg/internal/bzl_selects.bzl
roya1v: Thanks a lot for the idea!  I had to rely on `expr_members` being passed by reference to extract more logic and I feel like it makes the code a little bit harder to read, but I'm a starlark novice, so maybe I'm just not used to this behaviour being implicit. What do you think?

luispadron (approved):

swiftpkg/internal/bzl_selects.bzl
luispadron: Coming from swift it does feel a lil weird but I think that's pretty common in Starlark/python so this is looking better IMO, thank you!

cgrindel (approved):

cgrindel:
@mergifyio queue

PR #1176, opened 2024-07-17
README update

omarzl:
Hello!

I followed the instructions in the README, but I got this error:
```
ERROR: no such package '@@[unknown repo 'swift_deps_info' requested from @@]//': The repository '@@[unknown repo 'swift_deps_info' requested from @@]' could not be resolved: No repository visible as '@swift_deps_info' from main repository
```
After looking at the Example files, I found out that I was missing the line `"swift_deps_info"` in `use_repo`. `bazel mod tidy` was not generating it.

After adding the line `declare_swift_deps_info = True` it was generated and I was able to compile correctly, that is why I am updating the README file

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,7 @@ swift_deps = use_extension(
     "swift_deps",
 )
 swift_deps.from_package(
+    declare_swift_deps_info = True,
     resolved = "//:Package.resolved",
     swift = "//:Package.swift",
 )


cgrindel:
Thanks for the submission! You are correct that the `README.md` is not clear on when to enable to `declare_swift_deps_info`. I just put up #1185 providing more context. Can you take a look and provide feedback on that PR?

PR #1183, opened 2024-07-25
fix: Update bundle name generator to match SPM

jflan-dd:
SPM generated bundles appear to have the format `<package_name>_<module_name>.bundle` rather than `<module_name>_<module_name>.bundle`.

I added the same test with the same name to both the `resources_example` bazel tests and directly to the `CoolUI` example package to prove that the generated names match.


jflan-dd:
@cgrindel This is ready for review when you get a chance

cgrindel:
I am on vacation. I will take a look next week.

cgrindel (approved):
Thank you!

PR #1185, opened 2024-07-29
docs: clarify when to enable `declare_swift_deps_info`

cgrindel:
Clarify when the `declare_swift_deps_info` attribute should be enabled for workspaces that use  `rules_swift_package_manager`.

Closes #1176.


cgrindel:
cc: @omarzl

jpsim (approved):

Issue #1212, opened 2024-08-28
Bug: Package names not being applied correctly?


jpsim:
I just tried updating to the latest versions of [swift-composable-architecture](https://github.com/pointfreeco/swift-composable-architecture/releases/tag/1.14.0) and [swift-navigation](https://github.com/pointfreeco/swift-navigation/releases/tag/2.2.0) and am encountering this compiler error:

```
Sources/ComposableArchitecture/Effects/Cancellation.swift:191:50: error: ambiguous use of '_rethrowGet()'
189 |       return try await task.cancellableValue
190 |     } catch {
191 |       return try Result<T, Error>.failure(error)._rethrowGet()
    |                                                  `- error: ambiguous use of '_rethrowGet()'
192 |     }
193 |   }
    :
263 |   }
264 | 
265 |   func _rethrowGet() rethrows -> Output {
    |        `- note: found this candidate 
266 |     return try get()
267 |   }

Sources/SwiftNavigation/Internal/ErrorMechanism.swift:13:16: note: found this candidate in module 'SwiftNavigation'
11 |   }
12 | 
13 |   package func _rethrowGet() rethrows -> Output {
   |                `- note: found this candidate in module 'SwiftNavigation'
14 |     return try get()
15 |   }
```

This is odd because they have different package names, so a `package` ACL declaration from [swift-navigation](https://github.com/pointfreeco/swift-navigation/blob/2.2.0/Package%40swift-6.0.swift#L6) should not be visible to [swift-composable-architecture](https://github.com/pointfreeco/swift-composable-architecture/blob/1.14.0/Package.swift#L7).

jpsim:
For now I can work around this issue by applying a small patch to swift-navigation:

```diff
--- a/Sources/SwiftNavigation/Internal/ErrorMechanism.swift
+++ b/Sources/SwiftNavigation/Internal/ErrorMechanism.swift
@@ -10,6 +10,7 @@ extension _ErrorMechanism {
     fatalError()
   }
 
+  @_disfavoredOverload
   package func _rethrowGet() rethrows -> Output {
     return try get()
   }
```

cgrindel:
This is strange. Does this error occur when building a `swift_xxx` target that has a dependency on both `swift-composable-architecture` and `swift-navigation`? I presume that it does not happen when you build the targets in these external dependencies.

jpsim:
[...]

PR #1215, opened 2024-09-03
fix: use package name for `package_name` resolution

jpsim:
Fixes https://github.com/cgrindel/rules_swift_package_manager/issues/1212


luispadron (approved):
Thanks! This looks correct now going off the fixture diff. You'll need to update then, I think via `bazel run tidy`

jpsim:
`bazel run tidy` didn't do anything for me, but I was able to run `bazel test //swiftpkg/tests:swiftpkg_build_files_tests_test_0`, see the failure and fix the test manually.

PR #1223, opened 2024-09-14
fix: ignore unresolved packages in Package.swift

luispadron:
This fixes an issue where if a brand new dependency is added to the `Package.swift` when `Package.resolved` has not yet been updated, the Bazel build fails. It fails because there is no `pin` available for the `source_control` dependency as the `Package.resolved` has not yet been updated.

For example, a command line tool built using `rules_swift_package_manager` which has a `Package.swift` for `swift-argument-parser`. If a new dependency is added to the `Package.swift` but not yet used in any of the Bazel targets and no `swift package resolve` has been run when a build is run the build will fail with:

```
File bazel-path/external/rules_swift_package_manager~/swiftpkg/bzlmod/swift_deps.bzl", line 152, column 25, in _declare_pkg_from_dependency
                commit = pin.state.revision,
Error: 'NoneType' value has no field or method 'state'
```

This change allows the build to progress by simply ignoring the unresolved dependencies until a `swift package resolve` is run by the user.

### DIFF
--- a/swiftpkg/bzlmod/swift_deps.bzl
+++ b/swiftpkg/bzlmod/swift_deps.bzl
@@ -48,6 +48,19 @@ def _declare_pkgs_from_package(module_ctx, from_package, config_pkgs):
     direct_dep_repo_names = []
     direct_dep_pkg_infos = {}
     for dep in pkg_info.dependencies:
+        # Ignore unresolved dependencies, for example for a new packgage added
+        # to the `Package.swift` which has not been resolved yet.
+        # By ignoring these for now we can allow the build to progress while
+        # expecting a resolution in the future.
+        if not dep.file_system and \
+           not dep.source_control:
+            # buildifier: disable=print
+            print("""
+WARNING: {name} is unresolved and won't be available duing the build, resolve \
+the Swift package to make it available.\
+""".format(name = dep.name))
+            continue
+
         bazel_repo_name = bazel_repo_names.from_identity(dep.identity)
         direct_dep_repo_names.append(bazel_repo_name)
         pkg_info_label = "@{}//:pkg_info.json".format(bazel_repo_name)
@@ -168,9 +181,6 @@ def _declare_pkg_from_dependency(dep, config_pkg):
             path = dep.file_system.path,
             dependencies_index = None,
         )
-
-    else:
-        fail("Unrecognized dependency type for {}.".format(dep.identity))
 
 def _swift_deps_impl(module_ctx):
     config_pkgs = {}
--- a/swiftpkg/internal/pkginfos.bzl
+++ b/swiftpkg/internal/pkginfos.bzl
@@ -690,10 +690,7 @@ def _new_dependency(identity, name, source_control = None, file_system = None):
     Returns:
         A `struct` representing an external dependency.
     """
-    if not source_control and not file_system:
-        fail("""\
-A dependency must have either a source_control or file_system arg.\
-""")
+
     return struct(
         identity = identity,
         name = pkginfo_dependencies.normalize_name(name),
@@ -710,6 +707,9 @@ def _new_source_control(pin):
     Returns:
         A `struct` representing source control info for a dependency.
     """
+    if not pin:
+        return None
+
     return struct(
         pin = pin,
     )


swiftpkg/bzlmod/swift_deps.bzl
```diff
@@ -48,6 +48,13 @@ def _declare_pkgs_from_package(module_ctx, from_package, config_pkgs):
     direct_dep_repo_names = []
     direct_dep_pkg_infos = {}
     for dep in pkg_info.dependencies:
+        # Ignore unresolved dependencies, for example for a new packgage added
+        # to the `Package.swift` which has not been resolved yet.
+        # By ignoring these for now we can allow the build to progress while expecting a resolution in the future.
+        if not dep.file_system and \
+           not dep.source_control:
+            continue
```
cgrindel, line 56: Should we print something saying that we detected this scenario?

cgrindel (commented):
Can you explain what the benefit is to the user? If I understand correctly, this would only be useful if you add a dep to `Package.swift` and not use it in your code.

luispadron:
> Can you explain what the benefit is to the user? If I understand correctly, this would only be useful if you add a dep to `Package.swift` and not use it in your code.

For us we need this because we have a CLI tool which uses the Package.swift to define deps. This CLI tool also has a command for working with the Package.swift, basically we automate updates for the Package.swift itself via the the CLI tool. As it is today, if I add a new dependency and then try to run `our-cli package resolve` it fails to build

swiftpkg/bzlmod/swift_deps.bzl
luispadron: Yeah good call, this should be a warning probably. I can get that added. 

Would we want to pull in a copy of: https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/44b6f046d95b84933c1149fbf7f9d81fd4e32020/xcodeproj/internal/logging.bzl#L24 this way we can sort of match how Bazel warnings look

cgrindel: I am not familiar with the module. Is it smart enough to not add the colorization if it is not being output to a TTY?

luispadron: Hmm probably not it's just using print afaik

cgrindel: I don't think that we should litter CI logs with color codes, if we can avoid it. For now, let's just print something. Perhaps we can come up with a Starlark logging module that can detect when to colorize or at least be configured not to colorize. We can add it to `bazel-contrib`.

luispadron: Added a `print` warning here

cgrindel:
Looks like we need to run `bazel run //:tidy`.

swiftpkg/bzlmod/swift_deps.bzl
```diff
@@ -48,6 +48,17 @@ def _declare_pkgs_from_package(module_ctx, from_package, config_pkgs):
...
     direct_dep_pkg_infos = {}
     for dep in pkg_info.dependencies:
+        # Ignore unresolved dependencies, for example for a new packgage added
+        # to the `Package.swift` which has not been resolved yet.
+        # By ignoring these for now we can allow the build to progress while expecting a resolution in the future.
+        if not dep.file_system and \
+           not dep.source_control:
+            # buildifier: disable=print
+            print("""
+WARNING: {name} is unresolved and won't be available duing the build, resolve the Swift package to make it available.
```
cgrindel, line 58:
```suggestion
WARNING: {name} is unresolved and won't be available duing the build, resolve \
the Swift package to make it available.\
```

swiftpkg/bzlmod/swift_deps.bzl
```diff
@@ -48,6 +48,17 @@ def _declare_pkgs_from_package(module_ctx, from_package, config_pkgs):
     direct_dep_repo_names = []
     direct_dep_pkg_infos = {}
     for dep in pkg_info.dependencies:
+        # Ignore unresolved dependencies, for example for a new packgage added
+        # to the `Package.swift` which has not been resolved yet.
+        # By ignoring these for now we can allow the build to progress while expecting a resolution in the future.
```
cgrindel, line 53:
```suggestion
        # By ignoring these for now we can allow the build to progress while 
        # expecting a resolution in the future.
```

cgrindel (changes requested):

cgrindel (approved):

cgrindel:
@mergifyio queue

PR #1260, opened 2024-09-25
fix: usage of `copts` for `SWIFT_PACKAGE`

### LINKED ISSUE #1259: SWIFT_PACKAGE copt not correctly applied when importing Objective-C library
When a Swift target depends on an Objective-C target that uses `#if SWIFT_PACKAGE` we are currently only adding the `-DSWIFT_PACKAGE` to `copts` but we should also be adding `-Xcc -DSWIFT_PACKAGE` for the clang compilation step.

---

The following build file reproduces an issue, it shows a minimal Package.swift BUILD files after rules_swift_package_manager generates it:

```starlark
load("@build_bazel_rules_swift//swift:swift.bzl", "swift_library", "swift_library_group", "swift_binary")
load("@rules_swift_package_manager//swiftpkg/internal:generate_modulemap.bzl", "generate_modulemap")

swift_library_group(
  name = "Foo",
  deps = ["@swiftpkg_ios_svgkit//:SVGKit.rspm", "@swiftpkg_ios_svgkit//:SVGKit.rspm_modulemap"],
)

objc_library(
  name = "Foo.rspm",
  srcs = ["Foo.m"],
  enable_modules = True,
  module_name = "Foo",
  hdrs = ["Foo.h"],
  copts = [
    "-fblocks",
    "-fobjc-arc",
    "-fPIC",
    "-fmodule-name=Foo",
    "-DSWIFT_PACKAGE=1",
  ],
)

generate_modulemap(
  name = "Foo.rspm_modulemap",
  deps = [],
  hdrs = ["Foo.h"],
  module_name = "Foo",
  noop = False,
)

swift_library_group(
  name = "FooSwift",
  deps = [":FooSwift.rspm"],
)

swift_library(
  name = "FooSwift.rspm",
  tags = ["manual"],
  deps = [
    ":Foo.rspm",
    ":Foo.rspm_modulemap"
  ],
  copts = ["-DSWIFT_PACKAGE"],
  module_name = "FooSwift",
  srcs = ["Foo.swift"],
  always_include_developer_search_paths = True,
)

# Consumer example

swift_binary(
    name = "swift_package_copt_example",
    srcs = ["main.swift"],
    deps = [":FooSwift"],
)

```

With the following `Foo.h`

```c
#if SWIFT_PACKAGE
int foo(int x);
#else
int bar(int x); // should not be hit when building as a Swift package
#endif
```

When running:

```sh
bazel run swift_package_copt_example
```

The following error is shown:

```
Foo.swift:5:12: error: cannot find 'foo' in scope
    return foo(x)
           ^~~
```

This is because the `#if SWIFT_PACKAGE` case is not being hit. Fixing this requires updating the `copts` attribute of the `swift_library` to be: `copts = ["-DSWIFT_PACKAGE", "-Xcc", "-DSWIFT_PACKAGE"]`. 

Both are required because the first one is for the Swift compile and the second for the clang compile step.

luispadron:
This primarily does two changes:

- Passes `-Xcc -DSWIFT_PACKAGE` for the Swift library target generated `copts` attribute.
- Removes de-duplication in `bzl_selects.bzl`
  - Using `sets` enforces unique values for the fields which is problematic, for example: `copts = ["-DSWIFT_PACKAGE", "-Xcc", "-DSWIFT_PACKAGE"]` are valid copts but would get de-duplicated into `copts = ["-DSWIFT_PACKAGE", "-Xcc"]`.
  - We should not work-around duplicated `deps`, `copts`, etc. in the `BUILD` file generator. If issues arise during the build the `Package.swift` for the package should be fixed if it is a legitimate duplicate.

Fixes #1259

### DIFF
--- a/swiftpkg/internal/bzl_selects.bzl
+++ b/swiftpkg/internal/bzl_selects.bzl
@@ -102,6 +102,11 @@ def _new_from_target_dependency_condition(kind, labels, condition = None):
         for p in spm_platforms.supported(condition.platforms)
     ]
 
+    # Because `spm_platforms.label` may transform two different platforms into
+    # the same Bazel label (e.g. `macos` and `driverkit`), we need to uniquify
+    # the conditions.
+    conditions = sets.to_list(sets.make(conditions))
+
     return [
         _new(kind = kind, value = labels, condition = c)
         for c in conditions
@@ -155,34 +160,33 @@ def _to_starlark(values, kind_handlers = {}, mutually_inclusive = False):
     # dict whose keys are the conditions and the value is the value for the
     # condition.
     selects_by_kind = {}
-    no_condition_results = sets.make()
+    no_condition_results = []
 
     for v in values:
         v_type = type(v)
         if v_type != "struct":
             if v_type == "list":
-                no_condition_results = sets.union(no_condition_results, sets.make(v))
+                no_condition_results.extend(v)
             else:
-                sets.insert(no_condition_results, v)
+                no_condition_results.append(v)
             continue
 
         # We are assuming that the select will always result in a list.
         # Hence, we wrap the transformed value in a list.
         kind_handler = kind_handlers.get(v.kind, _noop_kind_handler)
-        tvs_set = sets.make(lists.flatten(kind_handler.transform(v.value)))
+        tvs = lists.flatten(kind_handler.transform(v.value))
         if v.condition != None:
             # Collect all of the values associted with a condition.
             select_dict = selects_by_kind.get(v.kind, {})
-            condition_values = select_dict.get(v.condition, sets.make())
-            condition_values = sets.union(condition_values, tvs_set)
+            condition_values = select_dict.get(v.condition, []) + tvs
             select_dict[v.condition] = condition_values
             selects_by_kind[v.kind] = select_dict
         else:
-            no_condition_results = sets.union(no_condition_results, tvs_set)
+            no_condition_results = no_condition_results + tvs
 
     expr_members = []
-    if sets.length(no_condition_results) > 0:
-        expr_members.append(sets.to_list(no_condition_results))
+    if len(no_condition_results) > 0:
+        expr_members.append(no_condition_results)
     for (kind, select_dict) in selects_by_kind.items():
         kind_handler = kind_handlers.get(kind, _noop_kind_handler)
         sorted_keys = sorted(select_dict.keys())
@@ -191,13 +195,13 @@ def _to_starlark(values, kind_handlers = {}, mutually_inclusive = False):
             # Generate multiple select expressions for each condition.
             for k in sorted_keys:
                 new_dict = {
-                    k: sets.to_list(select_dict[k]),
+                    k: select_dict[k],
                 }
                 _append_select(expr_members, kind_handler, new_dict)
         else:
             # Combine all conditions of the same kind into one select expression.
             new_dict = {
-                k: sets.to_list(select_dict[k])
+                k: select_dict[k]
                 for k in sorted_keys
             }
             _append_select(expr_members, kind_handler, new_dict)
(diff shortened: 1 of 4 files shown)


swiftpkg/internal/swiftpkg_build_files.bzl
```diff
@@ -106,6 +106,9 @@ def _swift_target_build_file(pkg_ctx, target):
         # SPM directive instructing the code to build as if a Swift package.
         # https://github.com/apple/swift-package-manager/blob/main/Documentation/Usage.md#packaging-legacy-code
         "-DSWIFT_PACKAGE",
+        # SPM directive instructing the code to build as if a Swift package for any clang modules.
+        "-Xcc",
+        "-DSWIFT_PACKAGE",
```
cgrindel, line 111: I am confused. When are we building a C/C++ file with the swift compiler? We have separate targets for Swift and C/C++ source code.

cgrindel (commented):
Is there a publicly visible Swift package that exhibits this behavior? If so, can we add it to one of the examples?

swiftpkg/internal/swiftpkg_build_files.bzl
luispadron: -Xcc passes a flag through to all C compiler invocations

swiftpkg/internal/swiftpkg_build_files.bzl
```diff
@@ -106,6 +106,9 @@ def _swift_target_build_file(pkg_ctx, target):
         # SPM directive instructing the code to build as if a Swift package.
         # https://github.com/apple/swift-package-manager/blob/main/Documentation/Usage.md#packaging-legacy-code
         "-DSWIFT_PACKAGE",
+        # SPM directive instructing the code to build as if a Swift package for any clang modules.
```
luispadron, line 109:
```suggestion
        # Define the SPM directive through all C compiler invocations
```

luispadron:
Added an example repo and example in: https://github.com/cgrindel/rules_swift_package_manager/pull/1261

swiftpkg/internal/swiftpkg_build_files.bzl
cgrindel: Sorry. Perhaps, I am missing something. We create `cc_xxx`, `objc_xxx`, and `swift_xxx` targets separating the source files into their respective targets. How is a C/C++ file being compiled by the Swift compiler? The `cc_xxx` and `objc_xxx` should be compiled with clang.

brentleyjones: When Swift depends on a clang module (modulemap and headers), with implicit modules it compiles that clang module. `-Xcc` is a way to pass compiler flags to that compilation.

luispadron:
I'll need to update this because we do want to dedupe some attrs (like deps) across configs/defaults but not copt. Going to look at making a new method for this

luispadron:
@cgrindel this should be ready now, I updated the PR description and the example in #1261 that can be added after this merges.

swiftpkg/internal/bzl_selects.bzl
```diff
@@ -155,34 +160,33 @@ def _to_starlark(values, kind_handlers = {}, mutually_inclusive = False):
     # dict whose keys are the conditions and the value is the value for the
     # condition.
     selects_by_kind = {}
-    no_condition_results = sets.make()
+    no_condition_results = []
```
cgrindel, line 163: Why did you move away from the set? Are we sure that duplicates are not possible?

cgrindel (commented):
Before we merge this, I added a question about the move away from using a set.

swiftpkg/internal/bzl_selects.bzl
cgrindel: I just re-read the description on the PR. I will approve.

cgrindel (approved):

PR #1261, opened 2024-09-26
chore: add example of SWIFT_PACKAGE defines

luispadron:
Showcases an issue with `SWIFT_PACKAGE` defines when consuming a Swift Package like: https://github.com/luispadron/swift-package-defines-example

The package above builds with `swift build` but fails when building through Bazel in the added example (without #1260)

```sh
cd examples/interesting_deps && bazel build print

ERROR: 
...
external/rules_swift_package_manager~~swift_deps~swiftpkg_swift_package_defines_example/Sources/FooSwift/Foo.swift:5:24: error: cannot find 'foo' in scope
3 | public func fooSwift() {
4 |     print("FooSwift")
5 |     print("ObjC foo: \(foo())")
  |                        `- error: cannot find 'foo' in scope
6 | }
7 | 
```


cgrindel (approved):
Thank you!

PR #1298, opened 2024-10-23
Repro for single factor auth error

Tilak1028-st:


cgrindel:
There were some compilation issues in your repro. I fixed them. All appears to be working properly from an `rules_swift_package_manager` perspective. I pushed my updates back to your branch.

PR #1317, opened 2024-10-30
feat: set a default `--security-path` for swift_package_tool

luispadron:
Add support for setting a custom `--security-path`. This also ensures the global path is not used.

### DIFF
--- a/examples/interesting_deps/MODULE.bazel
+++ b/examples/interesting_deps/MODULE.bazel
@@ -61,6 +61,7 @@ swift_deps.configure_swift_package(
     dependency_caching = False,
     manifest_cache = "none",
     manifest_caching = False,
+    security_path = "spm-security",
 )
 use_repo(
     swift_deps,
--- a/swiftpkg/internal/swift_package_tool_runner_template.sh
+++ b/swiftpkg/internal/swift_package_tool_runner_template.sh
@@ -26,6 +26,7 @@ cache_path="%(cache_path)s"
 enable_build_manifest_caching="%(enable_build_manifest_caching)s"
 enable_dependency_cache="%(enable_dependency_cache)s"
 manifest_cache="%(manifest_cache)s"
+security_path="%(security_path)s"
 
 # Construct dynamic arguments.
 args=()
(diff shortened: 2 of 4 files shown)


examples/interesting_deps/MODULE.bazel
```diff
@@ -58,6 +58,7 @@ swift_deps.from_package(
 swift_deps.configure_swift_package(
     build_path = "spm-build",
     cache_path = "spm-cache",
+    security_path = "spm-security",
     dependency_caching = False,
     manifest_cache = "none",
     manifest_caching = False,
```
brentleyjones, line 64:
```suggestion
    dependency_caching = False,
    manifest_cache = "none",
    manifest_caching = False,
    security_path = "spm-security",
```

swiftpkg/internal/swift_package_tool_runner_template.sh
```diff
@@ -49,6 +50,7 @@ args+=("--manifest-cache=$manifest_cache")
   --package-path "$package_path" \
   --build-path "$build_path" \
   --cache-path "$cache_path" \
+  --security-path "$security_path" \
```
brentleyjones, line 53:
```suggestion
  --build-path "$build_path" \
  --cache-path "$cache_path" \
  --package-path "$package_path" \
  --security-path "$security_path" \
```

brentleyjones (approved):

cgrindel (changes requested):
Can you ping me once you have rebased this PR?

luispadron:
> I am not familiar with when/why one would need to use `--security-path`. Can we update an example to use it?

The `--security-path` sets the path for the fingerprints of the manifests and artifacts. Without setting this SPM will use a global path over in `~/.swiftpm`. While i was testing the Swift Package Registry work (which ill have a PR up for soon) I ended up poisoning my global cache (my registry implementation returned a bad SHA during development) and found it incredibly confusing how to reset that global cache.

For users of `@swift_package` it makes sense IMO to keep this path local to the Bazel workspace (in this case within the rules runfiles). If folks want to use global caches they can instead not use `@swift_package`.

I added an example using `security_path` [over here](https://github.com/cgrindel/rules_swift_package_manager/pull/1317/files#diff-bf78cd67724fdd6664afbbd4bc41130a59b6d689eb62da862b047aac3223abec)

cgrindel:
> I added an example using security_path [over here](https://github.com/cgrindel/rules_swift_package_manager/pull/1317/files#diff-bf78cd67724fdd6664afbbd4bc41130a59b6d689eb62da862b047aac3223abec)

Sorry that I missed that. Thank you!

cgrindel (approved):
LGTM

PR #1318, opened 2024-10-30
feat: add support for Swift package registries

### LINKED ISSUE #1016: Swift registry support

luispadron:
Fixes #1016 

This adds initial support for consuming packages using a [Swift Package Registry](https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/PackageRegistry/PackageRegistryUsage.md).

The majority of the changes are in the new `registry_swift_package` rule which creates a repository to build the targets from a package downloaded via a registry. 

Notably, because the identity of the package is used to derive the repository name, packages used via a registry have the following label: `@swiftpkg_[org].[name]`, so for example `@swiftpkg_swift_collections` becomes `@swiftpkg_apple.swift_collections` if using the `swift-collections` package from the `apple` organization (as registries support multiple orgs).

The `registry_swift_package` rule is mostly just an updated version of #1043 (thanks @watt)

## Open questions

- [x] Should the label replace `.` with `_` in identity? E.g. `@swiftpkg_apple.swift_collections` vs. `@swiftpkg_apple_swift_collections`: **NO**

## Follow-up

- `rule_swift_package_manager` already has a bootstrapping issue, if there is no `Package.resolved` defined already then the Bazel repositories fail to generate. The addition of registry support adds another bootstrapping issue with the `registries.json` file (users will need to run `swift package-registry` commands to create it currently). We should address both of these by making bootstrapping a new project easier.

## Caveats

- There is no public Swift Package Registry (that I know of), to support testing this in CI I created a fake registry over at: https://github.com/luispadron/fake-swift-package-registry which is hosted at: https://fake-spm-registry.deno.dev
  - Example: https://fake-spm-registry.deno.dev/apple/swift-collections
  - I can maintain the fake, read-only registry until we have an alternative or it somehow starts costing me too much money 😅 

--- 

# Example

The following shows an example of a `swift_binary` target depending on an SPM dependency downloaded from a registry defined in `registries.json`.

## `Package.swift`

```swift
let package = Package(
    name: "swift_package_registry_example",
    dependencies: [
        .package(id: "apple.swift-collections", exact: "1.1.3"),
    ]
)

```

## `Package.resolved`

```json
{
  "pins" : [
    {
      "identity" : "apple.swift-collections",
      "kind" : "registry",
      "location" : "",
      "state" : {
        "version" : "1.1.3"
      }
    }
  ],
  "version" : 2
}
```

## `registries.json`

```json
{
  "authentication" : {

  },
  "registries" : {
    "[default]" : {
      "supportsAvailability" : false,
      "url" : "https://fake-spm-registry.deno.dev/"
    }
  },
  "version" : 1
}
```

## `MODULE.bazel`

```starlark
swift_deps = use_extension(
    "@rules_swift_package_manager//:extensions.bzl",
    "swift_deps",
)
swift_deps.from_package(
    registries = "//:registries.json",
    resolved = "//:Package.resolved",
    swift = "//:Package.swift",
)
swift_deps.configure_swift_package(
    replace_scm_with_registry = True,
)
use_repo(
    swift_deps,
    "swift_package",
    "swiftpkg_apple.swift_collections",
)
```

## `BUILD.bazel`

```starlark
swift_binary(
    name = "swift_package_registry_example",
    srcs = ["main.swift"],
    module_name = "swift_package_registry_example",
    deps = ["@swiftpkg_apple.swift_collections//:Collections"],
)
```


swiftpkg/internal/bazel_repo_names.bzl
```diff
@@ -17,6 +17,16 @@ def _from_identity(identity):
...
+
+    # TODO: Should we do this replacement for registry packages?
+    # This potentially conflicts with Swift Package Manager's `--use-registry-identity-for-scm` flag.
+    # Without this though, what would have been generated as `swiftpkg_swift_collections` is instead
+    # `swiftpkg_apple.swift_collections` which seems more confusing and build breaking if going from registry
+    # to non-registry.
+    registry_org_identity_split = identity.split(".")
+    if len(registry_org_identity_split) > 1:
+        identity = registry_org_identity_split[1]
+
```
brentleyjones, line 29: Could we run into collision issues if a package has the name that we are replacing into?

luispadron: Yes, but I believe that would be the point of `--use-registry-identity-for-scm/--replace-scm-with-registry` flags in SPM. Since this makes up the package identity I imagine a registry user would want to use these flags always to enforce all dependencies resolve using the registry versions (or fail to find things outside of it).

Maybe what we should do here is:

- If `--replace-scm-with-registry` (via some equivalent Bazel attr) are set in the `swift_deps.configure_swift_package` extension, then we do this replacement. Otherwise we don't.

This means you can end up with potentially these two labels:

```
swftpkg_swift_collections
swiftpk_apple.swift_collection
```

luispadron: Decided against doing this transformation and instead added a `replace_scm_with_registry` attr to the `configure_swift_package` tag which will set the `--replace-scm-with-registry` SPM flag.

This means labels for registries will generally look like:

```
@swiftpkg_[org].[name]
```

examples/swift_package_registry_example/registries.json
luispadron, line 1: @cgrindel I originally had this in `.swiftpm/configurations/registries.json` to mimic the basic example of `swift package-registry` outputting this file but `bazel test @@//examples:swift_package_registry_example_test_bazel_.bazelversion` was failing to copy the input file. I looked a bit and it seems like the `.scratch` directories that the rule operates under isn't copying the hidden files like `.swiftpm`

cgrindel: This is [the code](https://github.com/bazel-contrib/rules_bazel_integration_test/blob/dd36af23c4a7c8e97d91f22cc99504a19107c901/tools/create_scratch_dir.sh#L71) that should find the hidden files to copy to scratch. Is `.swiftpm` a directory by any chance?

luispadron: Yes it is, or it was. I've moved it into the root of the workspace instead to work around the issue for now

cgrindel: Sorry. Refresh problem. I am just seeing your responses.

luispadron: For users the swiftpm directory would be created when adding a registry via `swift package-registry set <url>`

luispadron:
Im testing in a real registry implementation now and realizing some issues, going to make some updates but the majority of the design is ready for revie

brentleyjones:
> I can maintain the fake, read-only registry until we have an alternative or it somehow starts costing me too much money

Can we spin up a local webserver as part of CI instead?

cgrindel:
> Should the label replace . with _ in identity? E.g. @swiftpkg_apple.swift_collections vs. @swiftpkg_apple_swift_collections

I like the underscores version. However, it is not a strong preference.

cgrindel:
> I can maintain the fake, read-only registry until we have an alternative or it somehow starts costing me too much money
> Can we spin up a local webserver as part of CI instead?

+1 to @brentleyjones suggestion. Again, I am not too familiar with Swift registry hosting. If it is static, can we use GitHub to host it? I presume the functionality just needs a URL.

examples/interesting_deps/MODULE.bazel
cgrindel, line 1: Why did we need to update the `interesting_deps` example>

swiftpkg/internal/swift_package_tool_attrs.bzl
cgrindel, line 1: Curious. Why did you move these attributes to their own file?

luispadron:
> > I can maintain the fake, read-only registry until we have an alternative or it somehow starts costing me too much money
> 
> 
> 
> Can we spin up a local webserver as part of CI instead?

I tried this originally but local host doesn't support https and so the bazel download stuff fails because of insecure connection.

luispadron:
> > I can maintain the fake, read-only registry until we have an alternative or it somehow starts costing me too much money
> 
> > Can we spin up a local webserver as part of CI instead?
> 
> 
> 
> +1 to @brentleyjones suggestion. Again, I am not too familiar with Swift registry hosting. If it is static, can we use GitHub to host it? I presume the functionality just needs a URL.

It is not static, it needs to be a web server which follows the registry spec.

swiftpkg/internal/swift_package_tool_attrs.bzl
luispadron: They're used in 3 different files to share the arguments across the repo rule and the rule

examples/swift_package_registry_example/registries.json
luispadron: Should [this line](https://github.com/bazel-contrib/rules_bazel_integration_test/blob/dd36af23c4a7c8e97d91f22cc99504a19107c901/tools/create_scratch_dir.sh#L71) be updated to ignore `-type` instead?

Fine with using this work around for now until we fix it upstream and can go back to a more "simple" example with the default `.swiftpm` directory

luispadron:
> Im testing in a real registry implementation now and realizing some issues, going to make some updates but the majority of the design is ready for review

Done! I fixed my fake registry implementation, the `.zip` needed to contain a single directory named `scope.name` where i had `name-version`

examples/swift_package_registry_example/registries.json
cgrindel: > Should [this line](https://github.com/bazel-contrib/rules_bazel_integration_test/blob/dd36af23c4a7c8e97d91f22cc99504a19107c901/tools/create_scratch_dir.sh#L71) be updated to ignore -type instead?

[Yes.](https://github.com/bazel-contrib/rules_bazel_integration_test/issues/390) However, I suspect it will morph into "everything but .git".

While the fix to `rules_bazel_integration_test` is easy, the problem is that `rules_swift_package_manager`  has an incompatibility with the latest version(s) of `rules_bazel_integration_test`. As part of my work migrating the gazelle plugin out of `rules_swift_package_manager`, I fell into a rabbit hole chasing down an integration test failure using the latest version of `rules_bazel_integration_test`. So, in addition to the simple fix, I will need to figure out what broke.

You may wonder why `rules_bazel_integration_test` is not automatically updated in `rules_swift_package_manager` MODULE.bazel file. We tell Renovate to ignore it so that we do not bump the version of other non-dev dependencies.

cgrindel (approved):

cgrindel:
@luispadron When you are ready to merge, if you don't want to babysit the merge, you can specify `@mergifyio queue` in a comment. This will add it to the merge queue that we are using.

luispadron:
> Should the label replace . with _ in identity? E.g. @swiftpkg_apple.swift_collections vs. @swiftpkg_apple_swift_collections

I think the answer is here is: no? Thoughts? Doing the replacement would require manipulating the identity in a fair amount of the rules and im not sure it gives us much (plus the label is still different than non registry so we don't gain anything from doing the replacement)

examples/swift_package_registry_example/registries.json
luispadron: ah gotcha we can wait for that stuff to be resolved fully then and keep the workaround here for now. It's really only an issue in that integration test context, it works for normal use-cases

examples/swift_package_registry_example/Package.swift
```diff
@@ -0,0 +1,10 @@
+// swift-tools-version: 5.7
+
+import PackageDescription
+
+let package = Package(
+    name: "swift_package_registry_example",
+    dependencies: [
+        .package(id: "apple.swift-collections", exact: "1.1.3"),
```
watt, line 8: I ran into problems at first with `--replace-scm-with-registry`, because Package.resolved will contain a scoped ID like `apple.swift-collections` but every Package.swift will still reference the unscoped `swift-collections`.

To support that, I made changes to `swiftpkg/internal/deps_indexes.bzl` to remove the scope in `_new_product_index_key`. Not sure that's the ideal fix, but it worked.

Anyway, that's definitely something I would suggest testing out.

watt (approved):
🎉

brentleyjones:
> I think the answer is here is: no?

Yeah, let's not do the replacement.

swiftpkg/internal/registry_swift_package.bzl
```diff
@@ -0,0 +1,215 @@
...
+        id.scope,
+        id.name,
+        "{version}.zip".format(version = version),
+    )
+    repository_ctx.download_and_extract(
+        headers = _REGISTRY_ZIP_ACCEPT_HEADER,
+        url = archive_url,
+        output = output,
+        sha256 = checksum,
+        stripPrefix = "{scope}.{name}".format(scope = id.scope, name = id.name),
```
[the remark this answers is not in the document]
luispadron: Oh rough yeah, I had been using `<name>-<version>` for the prefix in my fake registry but then our Artifactory implementation was using `scope.name`. 

I'll update this to work like you described since it looks like we cant use `strip_prefix` here

swiftpkg/internal/registry_swift_package.bzl
```diff
@@ -0,0 +1,215 @@
...
+        checksum = archive_checksum,
+        id = id,
+        output = ".",
+        repository_ctx = repository_ctx,
+        registry_url = registry_url,
+        version = version,
+    )
+
+    # Download the `Package.swift` file to the root of the source archive output.
+    _download_package_swift(
```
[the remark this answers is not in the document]
luispadron: Yeah true.. I figured this was the point of this endpoint but we should probably just use the archive content and save ourselves a download. I'll update.

swiftpkg/internal/swift_package_tool_attrs.bzl
```diff
@@ -0,0 +1,75 @@
...
+        default = ".security",
+    ),
+    "use_registry_identity_for_scm": attr.bool(
+        doc = """Look up source control dependencies in the registry and use \
+their registry identity when possible to help deduplicate across the two \
+origins.
+
+NOTE: `replace_scm_with_registry = True` implies \
+`use_registry_identity_for_scm = True`.
+""",
```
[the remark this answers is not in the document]
luispadron: Yeah sorry the note isn't clear, it's an error to pass both flags but thats because `--replace-scm-with-registry` already implies `--use-registry-identity-for-scm`. I'll update the doc comment here to note that both cannot be set at once though

swiftpkg/internal/swift_package_tool_attrs.bzl
```diff
@@ -0,0 +1,75 @@
...
+        default = ".build",
+    ),
+    "cache_path": attr.string(
+        doc = "The relative path within the runfiles tree for the shared Swift Package Manager cache directory.",
+        default = ".cache",
+    ),
+    "config_path": attr.string(
+        doc = "The relative path within the runfiles tree for the Swift Package Manager configuration directory.",
+        default = ".swiftpm",
+    ),
```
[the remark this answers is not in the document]
luispadron: > Not quite. By default, swift looks at both a shared config in ~/.swiftpm and a local project config at .swiftpm. The --config-path argument specifies a different shared config path (there is no argument to change the project config path).

Yes thats how I understand this to work as well, the `.swiftpm` in this rule isn't in the project tree though its part of the rules runfiles.

> Setting this to .swiftpm (the project config path) effectively means swift will only look at .swiftpm and ignore the shared path.

Maybe the name i picked is confusing but this `.swiftpm` directory is going to live in the rules runfiles, it does not mean a `.swiftpm` directory in the repo.

The goal of the `swift_package_tool` is to attempt to control where SPM looks for caches, configs, etc. By setting `--config-path` here SPM will use a `registries.json` file that exists in the `swift_package_tool` rules runfiles tree.

swiftpkg/internal/swift_package_tool_runner_template.sh
```diff
@@ -43,12 +47,26 @@ else
...
+if [ "$use_registry_identity_for_scm" = "true" ]; then
+  args+=("--use-registry-identity-for-scm")
+fi
+
 args+=("--manifest-cache=$manifest_cache")
 
+# If registries_json is set, symlink the `.json` file to the `config_path/configuration` directory.
+if [ -n "$registries_json" ]; then
+  mkdir -p "$config_path" && ln -sf "$(readlink -f "$registries_json")" "$config_path/registries.json"
+fi
```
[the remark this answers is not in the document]
luispadron: > Won't this blow away a real registries.json file if one exists?

No, this is symlinking the real file into the rules runfiles so that SPM uses that one from the rules `--config-path`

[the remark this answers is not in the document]
luispadron: > I think you're also missing the configuration directory in here.

Seems to work without one but i'll update to follow the normal `.swiftpm` layout

examples/swift_package_registry_example/Package.swift
luispadron: > To support that, I made changes to swiftpkg/internal/deps_indexes.bzl to remove the scope in _new_product_index_key. Not sure that's the ideal fix, but it worked.

Ooh could you elaborate a bit here, is this an issue with `swift package resolve`, etc or with the way the rules generate something. I dont actually know whats required in a normal SPM project to go from SCM -> Registry, do folks need to delete the Package.resolved and re-resolve to replace the SCM identities?

swiftpkg/internal/registry_swift_package.bzl
```diff
@@ -0,0 +1,215 @@
...
+    https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/PackageRegistry/Registry.md#44-download-source-archive
+    """
+    archive_url = paths.join(
+        registry_url,
+        id.scope,
+        id.name,
+        "{version}.zip".format(version = version),
+    )
+    repository_ctx.download_and_extract(
+        headers = _REGISTRY_ZIP_ACCEPT_HEADER,
```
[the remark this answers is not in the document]
brentleyjones: Personally I think having this specific feature labelled as requiring Bazel 7.1 is fine. If people don't use this feature, it won't before 7.1, right? If so, it's not breaking anyone and people just need to upgrade if they want to use the new feature.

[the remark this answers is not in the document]
cgrindel: I think @brentleyjones is right. Let's document that it is only available in Bazel 7.1 or later. Go for it.

examples/swift_package_registry_example/Package.swift
watt: Just an issue with the rules I think.

When you run `swift package resolve --replace-scm-with-registry` the resolver does an inline replacement in all dependencies using the URL. If you run with `--verbose` you'll see messages like this every time it does a replacement of a dependency.

```
info: 'swift-async-algorithms': swizzling 'https://github.com/apple/swift-collections.git' with registry dependency 'apple.swift-collections'.
```

Then Package.resolved gets rewritten with `apple.swift-collections` instead of `swift-collections`. Unfortunately it doesn't record any additional info about the swap. So RSPM is left to deal with packages that still reference a dependency on `swift-collections` but a Package.resolved that doesn't contain it.

I think there's at least a couple options to deal with it:

1. Assume it's OK to ignore the scope during resolution, and index products from `apple.swift-collections` as `swift-collections`. This is easy and worked for me, but it will probably fail if there's ever two products that differ only by scope, with identical product names and package names. I wonder if `swift package resolve` could even handle that.
2. Have RSPM do its own reverse lookup and swizzle dependencies, duping some of the behavior of `swift package resolve`. I'm not sure what that would entail internally.

swiftpkg/internal/swift_package_tool_runner_template.sh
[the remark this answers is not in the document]
luispadron: Actually, seems like `--config-path` defines the `configuration` directory path, it's just that by default it's nested within `.swiftpm`. Going to keep this as-is but rename the default `config_path` value to `config` so its not confusing

swiftpkg/internal/registry_swift_package.bzl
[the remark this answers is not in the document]
luispadron: Updated this! Used a similar approach except using `cp` and `rm` instead of `find`

examples/swift_package_registry_example/Package.swift
luispadron: This is the last action item I need before merging this. Any thoughts here @cgrindel @brentleyjones, I may go with option 1 to start unless we have other ideas here.

cgrindel: > it will probably fail if there's ever two products that differ only by scope, with identical product names and package names

Is this case when the client specifies the same Swift package twice?

In general, I am fine with option 1.

brentleyjones: I would prefer option 2 if we can do it. Can we at least look at what SPM is doing for it's "swizzle" and see if we can replicate it?

luispadron: Yeah, I think we'd need to pre-fetch all possible metadata and save it for use within the rules. The rules would then replace URL identities with registry identities.

We'd have a `replace_scm_with_registry` option in the `swift_deps.from_package` call.

luispadron: If that makes sense, let me know. I think we could do this as a follow up as what I have right now works for the basic use-case of "i want to use a registry and dont care if everything is downloaded from the registry"

cgrindel: My opinion hasn't changed. I am happy with option 1 for now.

luispadron: Let's go with that then and follow-up as needed

watt: > > it will probably fail if there's ever two products that differ only by scope, with identical product names and package names
> 
> Is this case when the client specifies the same Swift package twice?

Haven't tested this, but I think one failure mode here is having two dependencies that independently depend on different packages with the same name *and* a product with the same name. For example:

- the root package depends on package A and package B
- package A depends on `github.com/foo/something`, product named `Common`
- package B depends on `github.com/bar/something`, product named `Common`
- `foo.something` is in the registry

In this case option 1 will resolve both deps to `foo.something`.

But I think this is a rare edge case, and like the idea to follow-up with improvements later.

> I believe the way this works is that SPM makes requests to the registry server to get the metadata for a package which typically returns a github url, from there it can replace the github url with the registry ID.

Almost, but reversed; SwiftPM [queries the registry with each URL](https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/PackageRegistry/Registry.md#endpoint-5) to see if there's an associated registry package. Since we'll already have all the resolved registry IDs in Package.resolved, we could probably create a mapping by querying it in either direction.

luispadron: Apologies for the long pause here, I got time to finish up the implementation and I'm happy with it as a first pass now. I decided to do the same thing SPM does here, e.g. we will look up any available identifiers from the registry given the source control URL of a package.

luispadron:
@cgrindel @watt @brentleyjones The PR has been updated, I believe I have addressed all comments here if folks want to give it another review. The biggest change was proper support for `replace_scm_with_registry`.

I also added a [doc on using Swift package registries over here](https://github.com/cgrindel/rules_swift_package_manager/blob/00897e2f759a402d49cb85c3bbaf919448c83cbf/docs/swift_package_registry.md)

brentleyjones:
Firebase failure is unrelated to this change (cc: @cgrindel).

cgrindel:
> Firebase failure is unrelated to this change (cc: @cgrindel).

@brentleyjones That is a weird error. Is this an issue in `rules_xcodeproj`?

brentleyjones:
Yes. Seems that **rules_xcodeproj** doesn't support Bazel 8 yet, in particular it looks like the change from `~` to `+` in Bzlmod labels.

swiftpkg/bzlmod/swift_deps.bzl
```diff
@@ -218,8 +242,10 @@ def _swift_deps_impl(module_ctx):
             config_pkgs[config_pkg.name] = config_pkg
     config_swift_package = None
     for mod in module_ctx.modules:
-        for config_swift_package in mod.tags.configure_swift_package:
-            config_swift_package = config_swift_package
+        for config_swift_package_tag in mod.tags.configure_swift_package:
+            if config_swift_package:
+                fail("Expected only one `configure_swift_package` tag, but found multiple.")
```
cgrindel, line 247:
```suggestion
                fail("""\
Expected only one `configure_swift_package` tag, but found multiple.\
""")
```

swiftpkg/internal/swift_package_tool.bzl
```diff
@@ -24,10 +32,17 @@ def _swift_package_tool_impl(ctx):
     template_dict.add("%(package_path)s", package_path)
     template_dict.add("%(build_path)s", build_path)
     template_dict.add("%(cache_path)s", cache_path)
+    template_dict.add("%(config_path)s", config_path)
     template_dict.add("%(enable_build_manifest_caching)s", "true" if ctx.attr.manifest_caching else "false")
     template_dict.add("%(enable_dependency_cache)s", "true" if ctx.attr.dependency_caching else "false")
     template_dict.add("%(manifest_cache)s", ctx.attr.manifest_cache)
+    template_dict.add("%(registries_json)s", registries.short_path if registries else "")
+    template_dict.add("%(replace_scm_with_registry)s", "true" if ctx.attr.replace_scm_with_registry else "false")
```
cgrindel, line 40: Line length.

swiftpkg/internal/swift_package_tool_attrs.bzl
```diff
@@ -0,0 +1,72 @@
...
+If not using the `swift_package_tool` rules, the file must be in one of \
+Swift Package Manager's search paths or in the manually specified \
+`--config-path` directory.
+""",
+    ),
+}
+
+_swift_package_tool_config_attrs = {
+    "build_path": attr.string(
+        doc = "The relative path within the runfiles tree for the Swift Package Manager build directory.",
```
cgrindel, line 25: Please check the line length for this file.

cgrindel (commented):
Overall, LGTM. Most of my comments are related to formatting (e.g., line length).

cgrindel:
> Yes. Seems that **rules_xcodeproj** doesn't support Bazel 8 yet, in particular it looks like the change from `~` to `+` in Bzlmod labels.

Something is wrong. The `.bazelversion` for the repo is set to 7.4.1. 🤔 This seems to suggest that `rules_bazel_integration_test` is not using the correct version information.

watt (commented):
I attempted to test this locally with a Package.swift that has a dependency graph requiring the SCM replacement, and it didn't seem to work for me unfortunately. Haven't dug into exactly where it failed yet.

docs/swift_package_registry.md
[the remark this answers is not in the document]
luispadron: Bazel 6 was dropped support was dropped from most of the rules so i think were good now

swiftpkg/internal/registry_swift_package.bzl
```diff
@@ -0,0 +1,381 @@
...
+        if resource.get("name") == _SOURCE_ARCHIVE_NAME:
+            source_archive = resource
+            break
+
+    if not source_archive:
+        fail("No source archive was found in the package JSON: {package_json}".format(package_json = package_json))
+
+    return source_archive.get("checksum")
+
+def _get_resolved_pin_for_url(*, registry_url, repository_ctx, resolved_pkg_map, url):
```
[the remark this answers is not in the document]
luispadron: Potentially, I'll leave that for a follow-up though

luispadron:
@cgrindel i fixed most of the line space issues, there doesnt seem to be anything enforcing the 80 limit FWIW so we may want to get some formatters/linters setup to do this.

luispadron:
@mergifyio queue

cgrindel:
@mergifyio queue

PR #1323, opened 2024-11-05
feat: add support for `systemLibrary` targets 

### LINKED ISSUE #248: Support `system` target types
Looking for a popular package that leverages `system` targets (i.e, `systemLibrary`).

#235 was reported as a failure to support `system` targets. However, it turns out that the real issue was that targets were not being properly filtered out before build file generation. For instance, test targets and targets only associated with phantom products should not be built.

lucasmarcal-faire:
Fixes #248

### DIFF
--- a/swiftpkg/internal/pkginfos.bzl
+++ b/swiftpkg/internal/pkginfos.bzl
@@ -375,6 +375,18 @@ def _new_target_from_json_maps(
                 pkg_path = pkg_path,
                 sources = clang_src_info.explicit_srcs + clang_src_info.hdrs,
             )
+    elif module_type == module_types.system_library:
+        clang_src_info = _new_clang_src_info_from_sources(
+            repository_ctx = repository_ctx,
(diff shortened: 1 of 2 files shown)


cgrindel:
@lucasmarcal-faire LGTM so far. I have two requests:
1. Can you add a Swift package that has system libraries to one of the examples?
2. Can you add [a target generation unit test](https://github.com/cgrindel/rules_swift_package_manager/blob/8d4e7134bc13c0b857d83b04c933c93e79809313/swiftpkg/tests/swiftpkg_build_files_tests.bzl#L484)?

Thank you for your efforts on this! It is greatly appreciated.

luispadron:
@lucasmarcal-faire Any plans to pick this back up? Thanks again for getting the work started here!

lucasmarcal-faire:
> @lucasmarcal-faire Any plans to pick this back up? Thanks again for getting the work started here!

Going to rebase this PR and add tests today

PR #1360, opened 2024-11-26
fix: correct issues with modulemap generation

luispadron:
Fixes duplicate definition errors caused by custom/generated modulemap files.

### DIFF
--- a/swiftpkg/internal/swiftpkg_build_files.bzl
+++ b/swiftpkg/internal/swiftpkg_build_files.bzl
@@ -229,6 +229,7 @@ def _c_child_library(
         rule_kind,
         srcs,
         language_standard = None,
+        modulemap_path = None,
         res_copts = None):
     child_attrs = dicts.omit(attrs, ["data"])
     child_copts = list(attrs.get("copts", []))
@@ -238,6 +239,8 @@ def _c_child_library(
     if language_standard:
         child_copts.append("-std={}".format(language_standard))
     child_attrs["copts"] = child_copts
+    if modulemap_path:
+        child_attrs["module_map"] = modulemap_path
     return build_decls.new(
         rule_kind,
         name,
@@ -434,8 +437,8 @@ def _clang_target_build_file(repository_ctx, pkg_ctx, target):
         modulemap_target_name = pkginfo_targets.modulemap_label_name(bzl_target_name)
         noop_modulemap = clang_src_info.modulemap_path != None
         modulemap_attrs = {
-            "deps": bzl_selects.to_starlark(modulemap_deps),
-            "hdrs": clang_src_info.hdrs,
+            "deps": [] if noop_modulemap else bzl_selects.to_starlark(modulemap_deps),
+            "hdrs": [] if noop_modulemap else clang_src_info.hdrs,
             "module_name": target.c99name,
             "noop": noop_modulemap,
             "visibility": ["//:__subpackages__"],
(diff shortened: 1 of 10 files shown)


swiftpkg/internal/swiftpkg_build_files.bzl
```diff
@@ -435,7 +438,7 @@ def _clang_target_build_file(repository_ctx, pkg_ctx, target):
         noop_modulemap = clang_src_info.modulemap_path != None
         modulemap_attrs = {
             "deps": bzl_selects.to_starlark(modulemap_deps),
-            "hdrs": clang_src_info.hdrs,
+            "hdrs": [] if noop_modulemap else clang_src_info.hdrs,
```
luispadron, line 441: Not needed but noticed we were adding all the hdrs here even in the noop case which seems wasteful.

brentleyjones:
We should use an aspect hint instead (both for the empty case and the generated case): https://github.com/bazelbuild/rules_swift/blob/61e2c46dbfac6b9c11ecfb2c8e92f3e157ea9f27/swift/swift_clang_module_aspect.bzl#L221

luispadron:
> We should use an aspect hint instead (both for the empty case and the supplied case): https://github.com/bazelbuild/rules_swift/blob/61e2c46dbfac6b9c11ecfb2c8e92f3e157ea9f27/swift/swift_clang_module_aspect.bzl#L221

hmm, this doesn't seem to work. I believe because its failing during the `objc_library (parent) -> objc_library (child)` case. 

I've tried setting the child `aspect_hints` to: `@build_bazel_rules_swift//swift:no_module` and the parent to a custom `swift_interop_hint` with the `module_map` and `module_name`. Is there an equivalent hint for `objc_library`? I dont think the interop hint does anything when its part of `deps` of another `objc_library` right?

luispadron:
@brentleyjones I did update this to generate a `swift_interop_hint` for the parent so that we don't set `module_map` for it. For the child though I'm not sure how to get it working without setting `module_map`.

cgrindel (changes requested):
Can you remove [these lines](https://github.com/cgrindel/rules_swift_package_manager/blob/3f9d430f39c057cc6925e00b93b61e9f3cd31f18/examples/firebase_example/.bazelrc#L24-L35) from the firebase example and ensure that it works? Also, can you remove [this section from the FAQ](https://github.com/cgrindel/rules_swift_package_manager/blob/3f9d430f39c057cc6925e00b93b61e9f3cd31f18/docs/faq.md?plain=1#L97-L110)?

swiftpkg/internal/swiftpkg_build_files.bzl
```diff
@@ -1055,7 +1080,9 @@ def _new_for_license(pkg_info, license):
 
 # MARK: - Constants and API Definition
 
-swift_location = "@build_bazel_rules_swift//swift:swift.bzl"
+swift_repository = "@build_bazel_rules_swift"
+swift_location = "{rules_swift}//swift:swift.bzl".format(rules_swift = swift_repository)
+no_module_aspect_hint_label = "{rules_swift}//swift:no_module".format(rules_swift = swift_repository)
```
cgrindel, line 1085: The `no_module_aspect_hint_label` does not appear to be used. Can we revert these three lines?

luispadron: Oops leftover from testing Brentley's suggestion, removed these changes.

brentleyjones (approved):

PR #1376, opened 2024-12-04
fix: support binary artifacts in nested dirs

luispadron:
This brings the rules more in line with what SPM seems to support (though I couldn't find any spec on this specifically).

The current binary artifact generation logic in rules_swift_package_manager assumes the binary target is at most one level deep in the archive directory while [SPM seems to just find all `.xcframework` directories it can](https://github.com/swiftlang/swift-package-manager/blob/c26c12f54357fb7246c0bdbe3483105389f056b8/Sources/Workspace/Workspace%2BBinaryArtifacts.swift#L699-L724). This PR adds an example where the current behavior (e.g. `-max-depth 1`) causes the `.rspm` targets to not be generated for most of the products of the Yoti package.


cgrindel:
@luispadron Should we get the tests to green before I review?

luispadron:
@cgrindel the errors should be resolved now. I had to revert my change to fail if no xcframeworks are found, we seem to currently allow binary targets (e.g. pre-built cli tool) to not find any `artifact_infos`. 

We should discuss what should happen here when we dont find anything as we are currently silently not generating the rspm target but likely requires some more work

brentleyjones:
I've seen that happen when there are `objc_library` targets without `tags = ["manual"]`, which will cause them to get picked up and built as macOS.

cgrindel:
I know what's happening. The `ios_sim` example has a "clean" test. The examples test runner [checks for `set_up_clean_test`](https://github.com/cgrindel/rules_swift_package_manager/blob/d0fce68841a5871db3974a0accb487f476bb98cb/examples/test_runner.sh#L50-L55). If it sees this file, it runs it and then executes `do_test`, again. 

The clean tests exist to confirm that the gazelle plugin is generating targets properly when no build files exist. The Gazelle plugin does not support Apple-specific targets. Hence, it creates generic Swift targets. The generic Swift targets will execute on the host. Hence, the xcframework fails because it does not have a MacOS variant.

Since the gazelle plugin is moving out of this repo, it is safe to remove the `set_up_clean_test` for this example.

cgrindel:
FYI. I pushed this change to your branch.

.github/actions/set_up_macos/action.yml
```diff
@@ -14,7 +14,7 @@ runs:
   steps:
     - uses: maxim-lobanov/setup-xcode@v1
       with:
-        xcode-version: '15.0.1'
+        xcode-version: '15.2'
```
cgrindel, line 17: This snuck in there from an experiment. I am leaving it as I am upgrading the tests and runners in another PR.

cgrindel (approved):

cgrindel:
@mergifyio queue

PR #1385, opened 2024-12-06
fix: set `alwayslink = True`

brentleyjones:
This matches SPM’s behavior. It’s needed to create dynamic frameworks from rspm targets.


brentleyjones:
I'll wait until #1389 lands, since the test fixtures will be easier after that.

brentleyjones:
Draft just because I want to land the other PR. Ready for review.

cgrindel (approved):

PR #1387, opened 2024-12-06
fix: suppress modulemap target via swift interop

luispadron:
rules_swift will create a modulemap automatically unless the `SwiftInteropInfo` is suppressed. Since the `generate_modulemap` rule itself is a modulemap it should not have a modulemap generated for it.

Screenshot 2024-12-06 at 12 11 01 PM

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -24,7 +24,7 @@ bazel_dep(name = "platforms", version = "0.0.6")
 # rules_swift_package_manager _must_ be a runtime dependency of rules_swift_package_manager.
 bazel_dep(
     name = "rules_swift",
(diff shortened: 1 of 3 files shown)


brentleyjones (approved):

luispadron:
Had to upgrade **rules_swift** to `2.2.0`

brentleyjones (approved):

cgrindel:
I presume that this is fixing a problem for a specific Swift package. If so, could we add it to an example?

brentleyjones:
It doesn't fix an issue, it just removes an unneeded invalid modulemap.

cgrindel (approved):

PR #1389, opened 2024-12-06
fix: fix modulemaps to behave like SPM

brentleyjones:
* If a modulemap needs to be generated (i.e. one isn’t provided by the target), we always generate it instead of letting rules_swift generate it for Swift targets; this ensures we have consistent and non-duplicate modulemaps
* Generated modulemaps are now nested in a similar way to rules_swift, preventing them from being picked up accidentally
* We no longer generate a modulemap for Swift targets, and instead use the `swift.propagate_generated_module_map` feature
* We no longer create another target for the Objective-C resource module accessor; this aligns with how SPM does it and removes another target that could try to generate a modulemap
* The parent target for clang targets is now only an aggregator (i.e. it doesn’t have headers of its own), preventing it from complicating the build graph
* The parent target for clang targets includes the generated modulemap as a child dep, removing the need to generate a noop modulemap
* When detecting a custom modulemap, we only consider the public includes paths, preventing accidentally picking up errant modulemaps


swiftpkg/internal/swiftpkg_build_files.bzl
```diff
@@ -393,8 +392,77 @@ def _clang_target_build_file(repository_ctx, pkg_ctx, target):
...
+        build_decls.new(
+            kind = swift_kinds.interop_hint,
+            name = aspect_hint_target_name,
+            attrs = {
+                "module_map": hint_module_map,
+                "module_name": target.c99name,
+            },
+        ),
+    )
+    attrs["aspect_hints"] = [aspect_hint_target_name]
```
brentleyjones, line 461: We can't use `no_module` because it prevents rules_swift from generating module maps for transitive targets, which minimally breaks `apple_static_xcframework_import`.

swiftpkg/internal/swiftpkg_build_files.bzl
```diff
@@ -393,8 +360,77 @@ def _clang_target_build_file(repository_ctx, pkg_ctx, target):
...
+                attrs = {
+                    # We can't get a full picture of transitive dependencies
+                    # like rules_swift can (we would need to re-implement the
+                    # clang aspect). `use` declarations are only needed when
+                    # `-fmodules-decluse` is specified, so we are probably fine.
+                    #
+                    # Ideally long term the modulemap code in rules_swift can be
+                    # added to `objc_library` as well so we can stop generating
+                    # modulemaps entierly.
+                    "deps": [],
```
brentleyjones, line 395: SPM also doesn’t set `use` declarations, so I feel fine about this.

brentleyjones:
I still need to update fixtures, but this should be ready to review.

luispadron (approved):

brentleyjones:
Firebase failure is unrelated to this change (cc: @cgrindel).

cgrindel (approved):

examples/firebase_example/.bazelrc
```diff
@@ -20,6 +20,3 @@ build --cxxopt='-std=gnu++14'
 # Firebase SPM support requires `-ObjC` linker option.
 # https://github.com/firebase/firebase-ios-sdk/blob/master/SwiftPackageManager.md#requirements
 build --linkopt='-ObjC'
-
-# Use sandbox for rules_xcodeproj to avoid duplicate definition errors.
-build:rules_xcodeproj --spawn_strategy=remote,worker,sandboxed,local
```
luispadron, line 25: 🎉

PR #1401, opened 2024-12-14
fix: ensure the xcodeproj project generation uses the correct Bazel binary in the `firebase_example`

cgrindel:
The `rules_xcodeproj` project generator calls Bazel. By default, it looks for `bazel` on the `PATH`. When running under `rules_bazel_integration_test` this will pick up the default Bazel version on the system. If that is Bazelisk, it will pull down the latest Bazel. The fix is to explicitly tell `xcodeproj` which Bazel binary to use.

- Add `tools/bazel` to the `firebase_example`. It inspects environment variables to determine the Bazel version to use and writes `tools/bazel_for_xcodeproj` which executes Bazel commands against the correct Bazel binary.
- Use `sanbox` for `rules_xcodeproj` builds in firebase example. This avoids the duplicate definition error.
- Add `print_and_run` bash function to `examples/firebase_example/do_test` to ease debugging future issues.


examples/firebase_example/.bazelrc
```diff
@@ -20,3 +20,6 @@ build --cxxopt='-std=gnu++14'
 # Firebase SPM support requires `-ObjC` linker option.
 # https://github.com/firebase/firebase-ios-sdk/blob/master/SwiftPackageManager.md#requirements
 build --linkopt='-ObjC'
+
+# Use sandbox for rules_xcodeproj to avoid duplicate definition errors.
+build:rules_xcodeproj --spawn_strategy=remote,worker,sandboxed,local
```
luispadron, line 25: I thought we had fixed this already? Or will @brentleyjones  PR fix the remaining edge cases

examples/firebase_example/analytics/AnalyticsExample/BUILD.bazel
```diff
@@ -44,6 +44,9 @@ build_test(
 
 xcodeproj(
     name = "xcodeproj",
+    # The ./tools/bazel_for_xcodeproj is generated by tools/bazel. It should
+    # not be checked into source control.
+    bazel_path = "./tools/bazel_for_xcodeproj",
```
luispadron, line 49: 👍

examples/firebase_example/tools/bazel
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
+set -o errexit -o nounset -o pipefail
+
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" > /dev/null && pwd)"
+
+# Order of Bazel selection:
+# 1. BAZEL_REAL: Use Bazel specified by Bazelisk
+# 2. BIT_BAZEL_BINARY: Use Bazel specified by rules_bazel_integration_test.
+# 3: bazel: Try to find Bazel on the PATH
```
luispadron, line 10: Should this just fail? Might be clearer to enforce bazelisk usage if we require a specific Bazel version

luispadron (approved):

examples/firebase_example/.bazelrc
cgrindel: I hope @brentleyjones fix will be the real fix. Once I fixed the Bazel binary issue, I ran into the duplicate definition error.

examples/firebase_example/tools/bazel
cgrindel: > Might be clearer to enforce bazelisk usage if we require a specific Bazel version

Are we requiring a specific Bazel version? This PR is merely ensuring that `rules_xcodeproj` uses the Bazel that was used by any parent processes.

PR #1429, opened 2025-01-06
fix: Replicates SPM default behavior for publicHeadersPath: If this is nil, the directory is set to include

0xLucasMarcal:
# Match SPM's default behavior for public headers path

When SPM processes a target without an explicit `publicHeadersPath`, it follows these rules:
1. If `publicHeadersPath` is nil and an `include/` directory exists, use `include/` as the public headers path
2. This makes all headers in the `include/` directory visible to dependent targets

This PR replicates this canonical behavior by:
- Checking for the existence of `include/` directory when `publicHeadersPath` is not set
- Using `include/` as the public headers path if it exists
- Making all headers in `include/*.h` visible to clang targets

https://developer.apple.com/documentation/packagedescription/target/publicheaderspath

### DIFF
--- a/swiftpkg/internal/pkginfos.bzl
+++ b/swiftpkg/internal/pkginfos.bzl
@@ -1067,6 +1067,14 @@ def _new_clang_src_info_from_sources(
             paths.normalize(paths.join(abs_target_path, public_hdrs_path)),
         )
 
+    # If the Swift package manifest does not specify a public headers path,
+    # use the default "include" directory, if it exists.
+    # This copies the behavior of the canonical Swift Package Manager implementation.
+    # https://developer.apple.com/documentation/packagedescription/target/publicheaderspath
+    if public_hdrs_path == None:
+        if repository_files.path_exists(repository_ctx, paths.join(abs_target_path, "include")):
+            public_includes.append(paths.join(abs_target_path, "include"))
+
     # If the Swift package manifest has explicit source paths, respect them.
     # (Be sure to include any explicitly specified include directories.)
     # Otherwise, use all of the source files under the target path.


brentleyjones:
Does this fix a certain use case? If so, can we add or adjust an example to show it?

swiftpkg/internal/pkginfos.bzl
```diff
@@ -1067,6 +1067,14 @@ def _new_clang_src_info_from_sources(
...
         )
 
+    # If the Swift package manifest does not specify a public headers path,
+    # use the default "include" directory, if it exists.
+    # This copies the behavior of the canonical Swift Package Manager implementation.
+    # https://developer.apple.com/documentation/packagedescription/target/publicheaderspath
+    if public_hdrs_path == None:
+        if repository_files.path_exists(repository_ctx, paths.join(abs_target_path, "include")):
+            public_includes.append(paths.join(abs_target_path, "include"))
+
```
cgrindel, line 1077: So, we have support for magical public header directories [here](https://github.com/cgrindel/rules_swift_package_manager/blob/d7c6984741b7ca96fdf18cfb9f3ff1e3dbba141e/swiftpkg/internal/clang_files.bzl#L69-L94). Do you understand why this is not working for the tree-sitter case?

0xLucasMarcal: This search runs on the public headers list. Without this patch, the list was empty, so there were no headers to check if it was in a magical directory.

0xLucasMarcal: All the other Clang libraries that I used were modulemaps to expose the headers and define modules.

cgrindel: Sorry, I don't understand. If there are no header files in the `include` directory, why does the build fail?

0xLucasMarcal: Here's a small project with instructions to reproduce the failure: https://github.com/0xLucasMarcal/rules_spm_implicit_include_dir_issue

0xLucasMarcal: Basically what happens is that when the clang lib doesn’t have publicHeadersDir set and also doesn’t have a modulemap, the rules doesn’t reach the headers inside of include dir

cgrindel: Thank you for providing the repro! Let's add your repro package to the `pkg_manifest_minimal` example. That will be a good test.

cgrindel, line 1077:
```suggestion
    # If public includes were specified, use them.
    # Else if the Swift package manifest does not specify a public headers path,
    # use the default "include" directory, if it exists.
    # This copies the behavior of the canonical Swift Package Manager implementation.
    # https://developer.apple.com/documentation/packagedescription/target/publicheaderspath
    public_includes = []
    if public_hdrs_path != None:
        public_includes.append(
            paths.normalize(paths.join(abs_target_path, public_hdrs_path)),
        )
    elif repository_files.path_exists(
        repository_ctx,
        paths.join(abs_target_path, "include"),
    ):
        public_includes.append(paths.join(abs_target_path, "include"))

```

cgrindel: This suggested code combines the previous `if` statement with your new logic.

0xLucasMarcal: Nice!

0xLucasMarcal: Done

examples/pkg_manifest_minimal/Package.swift
```diff
@@ -8,6 +8,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-argument-parser", from: "1.5.0"),
         .package(url: "https://github.com/apple/swift-log", from: "1.6.2"),
         .package(url: "https://github.com/nicklockwood/SwiftFormat", from: "0.55.4"),
+
```
0xLucasMarcal, line 11:
```suggestion
```

luispadron (approved):

examples/pkg_manifest_minimal/Sources/MyExecutable/MyExecutable.swift
```diff
@@ -10,6 +11,8 @@ struct MyExecutable: AsyncParsableCommand {
         print(namedGreeting.value)
 
         let complexClass = ComplexClass(name: "Olivia", age: 30, favoriteColors: ["blue"])
-        print(complexClass.greet())
+        complexClass.greet()
+        let farewellMessage = FarewellFramework.myclang_get_farewell_message(MYCLANG_FAREWELL_GOODBYE)
+        print(String(cString: farewellMessage!))
```
cgrindel, line 16: Why not print the output of the `complexClass.greet()` and the `farewellMessage`?

examples/pkg_manifest_minimal/do_test
```diff
@@ -9,10 +9,13 @@ script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" > /dev/null && pwd)"
...
-  local pattern=${1}
+  local expected=${1}
   local actual="${2}"
-  local err_msg="Expected to match. pattern: ${pattern}, actual: ${actual}"
-  [[ "${actual}" =~ ${pattern} ]] || (echo >&2 "${err_msg}" && exit 1)
+  local err_msg="Expected to match. expected: ${expected}, actual: ${actual}"
+  if [[ "${expected}" != "${actual}" ]]; then
+    echo >&2 "${err_msg}"
+    exit 1
+  fi
```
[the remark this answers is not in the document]
cgrindel: If the pattern check was failing, we should understand why it was failing. The test is using Bash regex.

swiftpkg/tests/swiftpkg_build_files_tests.bzl
```diff
@@ -189,7 +189,6 @@ _pkg_info = pkginfos.new(
             exclude_paths = [
                 "src/do_not_include_me.cc",
             ],
-            public_hdrs_path = "include",
```
cgrindel, line 192: Can you explain why this disappears for this test case?

cgrindel (changes requested):
Most of my comments are around the example and the affect on a unit test.

swiftpkg/tests/swiftpkg_build_files_tests.bzl
0xLucasMarcal: This PR adds include to `public_hdrs_path` when it's not set, making this change, we also test this behavior

examples/pkg_manifest_minimal/do_test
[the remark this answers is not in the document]
0xLucasMarcal: I don't have context on that. The other tests in this file aren't running regex match asserts, they use full strings, so I thought the best decision was to keep things simple.

examples/pkg_manifest_minimal/Sources/MyExecutable/MyExecutable.swift
0xLucasMarcal: `complexClass.greet()` already runs print, so this print in MyExecutable.swift just outputs `()`. Because the assert function of the test was broken, we haven't caught that before.

examples/pkg_manifest_minimal/do_test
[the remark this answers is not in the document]
cgrindel: I have put up #1458 to fix the existing issues with the test. Please revert your changes, removing the `assert_match` in this PR.

examples/pkg_manifest_minimal/Sources/MyExecutable/MyExecutable.swift
cgrindel: Fixed in #1458. Thanks for pointing that out.

cgrindel:
We will wait for the #1458 to merge, then rebase this PR on those changes.

examples/pkg_manifest_minimal/do_test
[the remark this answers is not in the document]
0xLucasMarcal: Nice!!! Let me revert my changes

cgrindel (changes requested):
It looks like you need to execute `bazel run //:tidy` to update some source files. Also, it looks like some integration tests are failing. Let me know when they are green and I will review.

PR #1440, opened 2025-01-13
fix: Ignore invalid xcframework paths

luispadron:
The rules should ignore invalid `.xcframework` paths like SPM does ([see here](https://github.com/swiftlang/swift-package-manager/blob/c26c12f54357fb7246c0bdbe3483105389f056b8/Sources/Workspace/Workspace%2BBinaryArtifacts.swift#L771-L780)). Unlike SPM however, in Starlark we do not have a readily available `.plist` decoder so we will simply ensure that an `Info.plist` file exists and let the build rules handle potential issues.

This fixes cases where I was getting warnings because our .zip file contains a `__MACOSX` directory that contains a `.xcframework` directory (this is used for metadata by macOS). All the files are prefixed with `._` so the Info.plist check here fixes including these bad paths in our search.

```
DEBUG: /private/var/tmp/_bazel_lpadron/9ae9416857eb79bb978de35a53d54970/external/rules_swift_package_manager~/swiftpkg/internal/repo_rules.bzl:156:14: WARNING: Found multiple XCFramework binary artifacts in the downloaded artifact: ["remote/archive/AppsFlyerLib-Dynamic.xcframework.zip/__MACOSX/AppsFlyerLib.xcframework", "remote/archive/AppsFlyerLib-Dynamic.xcframework.zip/AppsFlyerLib.xcframework"], using the last one.
```

### DIFF
--- a/swiftpkg/internal/repo_rules.bzl
+++ b/swiftpkg/internal/repo_rules.bzl
@@ -139,6 +139,17 @@ def _artifact_infos_from_path(repository_ctx, path):
             by_name = "*.xcframework",
         )
 
+    # NOTE: SPM validates the `.xcframework` paths by decoding the `Info.plist` file.
+    # This would be more involved to do in Starlark, so instead we'll assume
+    # that a `.xcframework` dir is a potential candidate if it contains a
+    # `Info.plist` file without checking the file contents.
+    # See: https://github.com/swiftlang/swift-package-manager/blob/c26c12f54357fb7246c0bdbe3483105389f056b8/Sources/Workspace/Workspace%2BBinaryArtifacts.swift#L771-L780
+    xcframework_dirs = [
+        xf
+        for xf in xcframework_dirs
+        if repository_files.path_exists(repository_ctx, paths.join(xf, "Info.plist"))
+    ]
+
     # If multiple found, use the last one which is what SPM currently does:
     # https://github.com/swiftlang/swift-package-manager/blob/c26c12f54357fb7246c0bdbe3483105389f056b8/Sources/Workspace/Workspace%2BBinaryArtifacts.swift#L699-L723
     if len(xcframework_dirs) > 1:


swiftpkg/internal/repo_rules.bzl
```diff
@@ -139,6 +140,15 @@ def _artifact_infos_from_path(repository_ctx, path):
             by_name = "*.xcframework",
         )
 
+    # Ignore some bad xcframework dirs, for example, when creating archives on macOS it's possible a `__MACOSX`
+    # dir is created, in this case, we should ignore that path as there should be a real xcframework dir elsewhere.
```
cgrindel, line 144: Can you elaborate on this? I don't understand when this happens. At first blush, it seems strange to me to ignore this directory out of hand.

luispadron: This has some info on `__MACOSX` specifically: https://superuser.com/questions/104500/what-is-macosx-folder-i-keep-seeing-in-zip-files-made-by-people-on-os-x but effectively its some folder that macOS / Finder will make automatically under certain cases. This happens when you right click > compress on macOS, but not when using tools like `zip`. The `__MACOSX` directory contains symlinks to the real file so what happens in this code is it find two xcframework dirs and provides a warning.

I agree that this is really an issue with the archive and it should be fixed there, but I expect there to be many of these archives in the wild so ignoring paths with that directory seemed like a better alternative.

cgrindel: Does SPM warn/error due to this extra directory?

luispadron: It does warn for when it find multiple dirs, but looks like that case wouldnt be hit for `__MACOSX` as it's a symlink and that gets filtered out by the `isDirectory` check: https://github.com/swiftlang/swift-package-manager/blob/c26c12f54357fb7246c0bdbe3483105389f056b8/Sources/Workspace/Workspace%2BBinaryArtifacts.swift#L736C34-L736C45

So we could do the same here (only check directories, ignore symlinks, etc), let me know your thoughts

cgrindel: That feels better to me.

luispadron: So I misunderstood how this worked, `__MACOSX` aren't symlinks, they're actual directories with `._` prefixed files, see this output after running `unzip` on: https://github.com/AppsFlyerSDK/AppsFlyerFramework/releases/download/6.11.2/AppsFlyerLib.xcframework.zip

```
AppsFlyerLib
├── AppsFlyerLib.xcframework
│   ├── Info.plist
│   ├── ios-arm64_armv7
│   │   └── AppsFlyerLib.framework
│   ├── ios-arm64_i386_x86_64-simulator
│   │   └── AppsFlyerLib.framework
│   ├── ios-arm64_x86_64-maccatalyst
│   │   └── AppsFlyerLib.framework
│   ├── macos-arm64_x86_64
│   │   └── AppsFlyerLib.framework
│   ├── tvos-arm64
│   │   └── AppsFlyerLib.framework
│   └── tvos-arm64_x86_64-simulator
│       └── AppsFlyerLib.framework
└── __MACOSX
    └── AppsFlyerLib.xcframework
        ├── ._Info.plist
        ├── ._ios-arm64_armv7
        ├── ._ios-arm64_i386_x86_64-simulator
        ├── ._ios-arm64_x86_64-maccatalyst
        ├── ._macos-arm64_x86_64
        ├── ._tvos-arm64
        ├── ._tvos-arm64_x86_64-simulator
        ├── ios-arm64_armv7
        ├── ios-arm64_i386_x86_64-simulator
        ├── ios-arm64_x86_64-maccatalyst
        ├── macos-arm64_x86_64
        ├── tvos-arm64
        └── tvos-arm64_x86_64-simulator
```
In the Bazel rules, because we are using `.xcframework` extension to assume something is a valid XCFramework then we end up finding these two paths.


I'm not entirely sure how SPM works around these files (or how it finds valid .xcframeworks). Going to see if I can find out what they're using to find the right xcframework path(s) here.

luispadron: Looks like they're reading the `Info.plist` file to figure this out: https://github.com/swiftlang/swift-package-manager/blob/c26c12f54357fb7246c0bdbe3483105389f056b8/Sources/Workspace/Workspace%2BBinaryArtifacts.swift#L771-L780

🙃 this would a bit more involved to do in Starlark though

luispadron: Going with a simpler check of `Info.plist` existence. Updated the PR and description.

cgrindel (approved):

cgrindel:
@mergifyio queue

PR #1446, opened 2025-01-15
chore: add CODEOWNERS file

luispadron:
Adds the `CODEOWNERS` file to the repo, we may need to update the repo settings so that reviewers are requested automatically.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+* @brentleyjones @cgrindel @jpsim @luispadron


cgrindel (approved):

cgrindel:
@mergifyio queue

cgrindel:
I updated the branch protection rules to require a CODEOWNER review. For this repo, I don't think it matters too much, because we do not segregate different areas of the codebase by owner. If we think it is causing a problem, I will disable it.

PR #1452, opened 2025-01-20
Remove duplicated srcs

karim-alweheshy:
This resolves a case for using https://github.com/facebook/yoga from https://github.com/layoutBox/FlexLayout

### DIFF
--- a/swiftpkg/internal/clang_files.bzl
+++ b/swiftpkg/internal/clang_files.bzl
@@ -250,7 +250,7 @@ def _collect_files(
 
     modulemap = None
     modulemap_orig_path = None
-    for orig_path in all_srcs:
+    for orig_path in depset(all_srcs).to_list():
         path = _relativize(orig_path, relative_to)
         _root, ext = paths.split_extension(path)
         if lists.contains(_HEADER_EXTS, ext):


swiftpkg/internal/clang_files.bzl
```diff
@@ -250,7 +250,7 @@ def _collect_files(
 
     modulemap = None
     modulemap_orig_path = None
-    for orig_path in all_srcs:
+    for orig_path in depset(all_srcs).to_list():
```
cgrindel, line 253: Can you elaborate how we are getting duplicate sources in the first place?

cgrindel (commented):
Can we add the package that exhibits a problem to one of the examples?

swiftpkg/internal/clang_files.bzl
karim-alweheshy: [This](https://github.com/facebook/yoga/pull/1779) needs to be merged first and then I'll update this PR

cgrindel: Will you even need this PR, once the other PR merges?

karim-alweheshy: yes the other PR fixes something different related to picking up public headers unrelated to the compiled target

examples/yoga/README.md
cgrindel, line 1: Thank you for adding the example. To fully integrate the example with the repo, you will need to add the example to [this list](https://github.com/cgrindel/rules_swift_package_manager/blob/c320011d65d4fcbe75064c51d5014440c63280e5/examples/example_infos.bzl#L134-L152). Then, you will need to run `bazel run //:tidy`. This will update some source files that will need to be added to this PR.

PR #1458, opened 2025-01-25
fix: `pkg_manifest_minimal` example

cgrindel:
As pointed out #1429, the `pkg_manifest_minimal` example was calling `ComplexClass.greet()` and printing the result, but the function did not return anything. Instead, it was printing in `greet()` function. That was not the original intent. I updated `ComplexClass.greet()` to return the greeting so that the executable can print it. I also added a check in `do_test` to confirm that it was being output properly. All of this was a miss on my part.

Thanks to @0xLucasMarcal  for pointing this.


luispadron (approved):

cgrindel:
@mergifyio queue

PR #1039, opened 2024-04-24
fix: don't include directories in resources

jflan-dd:
I'm not sure if this is the best solution, but it did solve my issues. Happy to discuss other approaches.

This is the error I'm seeing:
```
Traceback (most recent call last):
	File "/private/var/tmp/_bazel_johnflanagan/c6ca44c5b5b4e1c56873b1d5235fffed/external/rules_apple~3.5.1/apple/internal/ios_rules.bzl", line 444, column 41, in _ios_application_impl
		processor_result = processor.process(
	File "/private/var/tmp/_bazel_johnflanagan/c6ca44c5b5b4e1c56873b1d5235fffed/external/rules_apple~3.5.1/apple/internal/processor.bzl", line 747, column 36, in _process
		partial_outputs = [partial.call(p) for p in partials]
	File "/private/var/tmp/_bazel_johnflanagan/c6ca44c5b5b4e1c56873b1d5235fffed/external/bazel_skylib~1.5.0/lib/partial.bzl", line 43, column 28, in _call
		return partial.function(*function_args, **function_kwargs)
	File "/private/var/tmp/_bazel_johnflanagan/c6ca44c5b5b4e1c56873b1d5235fffed/external/rules_apple~3.5.1/apple/internal/partials/resources.bzl", line 316, column 26, in _resources_partial_impl
		resources.deduplicate(
	File "/private/var/tmp/_bazel_johnflanagan/c6ca44c5b5b4e1c56873b1d5235fffed/external/rules_apple~3.5.1/apple/internal/resources.bzl", line 962, column 22, in _deduplicate
		field_handler(field, deduplicated)
	File "/private/var/tmp/_bazel_johnflanagan/c6ca44c5b5b4e1c56873b1d5235fffed/external/rules_apple~3.5.1/apple/internal/partials/resources.bzl", line 308, column 37, in _deduplicated_field_handler
		result = processing_func(**processing_args)
	File "/private/var/tmp/_bazel_johnflanagan/c6ca44c5b5b4e1c56873b1d5235fffed/external/rules_apple~3.5.1/apple/internal/partials/support/resources_support.bzl", line 261, column 53, in _datamodels
		datamodel_groups.update(group_files_by_directory(
	File "/private/var/tmp/_bazel_johnflanagan/c6ca44c5b5b4e1c56873b1d5235fffed/external/rules_apple~3.5.1/apple/utils.bzl", line 117, column 13, in group_files_by_directory
		fail("Expected only files inside directories named with the extensions " +
[...]

PR #1488, opened 2025-02-13
fix: Optionally exclude directories from list_files_under for performance

jflan-dd:
This PR adds an `exclude_directories` argument to `repository_files.list_files_under` which resulted in a massive performance improvement for packages with lots of resources.

I noticed that it was taking a really long time to "fetch" a package that was local to my machine, so there was no actual network operations required.

I dug into the issue and it's actually from a change I added in [this PR](https://github.com/cgrindel/rules_swift_package_manager/pull/1039) to expand certain types of resources into a list of their assets. The overwhelming majority of the time was being spent on the `if not repository_files.is_directory(repository_ctx, p)` check inside of the loop. By omitting directories from the result of `list_files_under` we can skip all of that work.

### Before - 225s
image

### After - 2s
image

I didn't see any uses of `list_files_under` that were relying on it to contain directories. Should we change the behavior to never include directories?

### DIFF
--- a/swiftpkg/internal/pkginfos.bzl
+++ b/swiftpkg/internal/pkginfos.bzl
@@ -309,15 +309,12 @@ def _new_target_from_json_maps(
     for directory in resource_directories:
         sets.remove(resources_set, directory)
 
-        resource_files = [
-            p
-            for p in repository_files.list_files_under(
-                repository_ctx,
-                directory.path,
-                exclude_paths = exclude_paths,
-            )
-            if not repository_files.is_directory(repository_ctx, p)
-        ]
+        resource_files = repository_files.list_files_under(
+            repository_ctx,
+            directory.path,
+            exclude_paths = exclude_paths,
+            exclude_directories = True,
+        )
 
         for p in resource_files:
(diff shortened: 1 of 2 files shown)


luispadron (approved):
Great find, thank you!

jflan-dd:
@luispadron Do you have any thoughts on just changing the behavior of `list_files_under` to always exclude directories?

It looked like all of the current uses would be safe without directories, and it would match the name of the function better.

luispadron:
> @luispadron Do you have any thoughts on just changing the behavior of `list_files_under` to always exclude directories?
> 
> 
> 
> It looked like all of the current uses would be safe without directories, and it would match the name of the function better.

Yeah that might make more sense, I'd try and see if the tests pass. @cgrindel thoughts?

jflan-dd:
> I'd try and see if the tests pass

Tests passed and my build still worked

luispadron:
I'll wait for chuck to have a chance to review but this looks great, thanks!

cgrindel (approved):

cgrindel:
Thank you!

cgrindel:
@mergifyio queue

PR #1489, opened 2025-02-13
fix: watch local packages to refetch if files are added/updated

jflan-dd:
Call `repository_ctx.watch_tree` to ensure that repositories are refetched for local SPM packages if new files are added or removed. This will also cause a refetch if a file's content changes, but I didn't see a way to only watch for changes to file creation/deletion.

Without this I was experiencing issues where local SPM packages would fail to compile after pulling down changes from git because new files were added without the repo's `BUILD.bazel` file being updated as well.

### DIFF
--- a/swiftpkg/bzlmod/swift_deps.bzl
+++ b/swiftpkg/bzlmod/swift_deps.bzl
@@ -60,7 +60,7 @@ def _declare_pkgs_from_package(module_ctx, from_package, config_pkgs, config_swi
            not dep.source_control:
             # buildifier: disable=print
             print("""
-WARNING: {name} is unresolved and won't be available duing the build, resolve \
+WARNING: {name} is unresolved and won't be available during the build, resolve \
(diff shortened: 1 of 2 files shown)


luispadron (approved):

cgrindel (commented):
[The watch_tree function was introduced in Feb 2024](https://github.com/bazelbuild/bazel/commit/9a4b712de12175dbf62d3e80665a68d07dcd2046). I am not exactly sure which Bazel version introduced this function. It was some time after Bazel 7.0.2.

Is it possible to check if `repository_ctx` has the function before attempting to call it?

jflan-dd:
@cgrindel Good catch! It looks like [it was added in 7.1.0](https://github.com/bazelbuild/bazel/pull/21435). I added a check to make sure the function exists before calling it

cgrindel (approved):
Thank you!

cgrindel:
@mergifyio queue

PR #1516, opened 2025-03-11
fix: Add custom module maps to `hdrs`

brentleyjones:
We add the custom module map to `hdrs` to insure it's part of the inputs of downstream targets. Without this sandboxed or RBE builds can fail.


luispadron (approved):

cgrindel (commented):
Is there something that we can add to an example to exercise this?

brentleyjones:
Could have something depend on `OCMock` and compile with sandbox enabled.

brentleyjones:
I think this is valuable even without an example.

PR #1521, opened 2025-03-13
fix: pass registry flags when dumping & describing

luispadron:
In the example we failed to test when a "mixed" `Package.swift` uses both `id:` and `url:` for `package`. In the mixed case, `swift_package` is generated for the `url:` declarations. This rule was not using the required registry arguments to make the identity replacement possible.

This was resulting in errors where if a registry package like `apple.swift-collections` was defined and a git package was also defined like `swift-async-algorithms` (which depends on `swift-collections`) the resulting labels in deps would not be the proper registry id e.g. `@swiftpkg_swift_collections` instead of `@swiftpkg_apple.swift_collections`.

This PR fixes the issues by passing down the required flags to the `swift package dump-package` and `swift package describe` commands that are run. 

This is now fixed for both `swift_package` and `registry_swift_package` (which removes a ton of unnecessary replacement logic that the dump/describe commands handle for us instead)


cgrindel (approved):

brentleyjones (approved):

luispadron:
One thing i need to follow-up on is that SPM does not seem to error out if for example the registry is down or you fail to reach it. In this case the dump-package/describe commands do not have the identities correctly swizzled for their registry version

luispadron:
@Mergifyio queue

PR #1542, opened 2025-03-24
fix: rules_apple in resources example

cgrindel:
The resources example started failing after upgrading to `rules_apple` 3.20.0. @luispadron suggestion of increasing the `destination_timeout` seems to be the workaround.

Continue investigation started in #1529.


cgrindel:
@mergifyio queue

PR #1556, opened 2025-03-27
feat: Use higher level GRPC api and remove forked convenience wrappers

AttilaTheFun:
I was previously looking at some old examples in the grpc-swift repo and didn't notice there were higher-level apis I could use which didn't require the wrappers. 

We can clean these up now.


cgrindel (approved):

cgrindel:
@mergifyio queue

cgrindel:
@mergifyio queue

PR #1559, opened 2025-03-28
fix: resource bundle accessor with swift concurrency + swift 6 mode

maxwellE:
In swift 6 mode + swift concurrency enabled the logic today cannot build with the following error:

```terminal
error: static property 'module' is not concurrency-safe because it is nonisolated global shared mutable state
 9 |     /// This is a workaround for the provided module static var that doesn't always work.
10 |     /// https://forums.swift.org/t/unable-to-find-bundle-in-package-target-tests-when-package-depends-on-another-package-containing-resources-accessed-via-bundle-module/43974/5
11 |     static var module: Bundle = {
   |                |- error: static property 'module' is not concurrency-safe because it is nonisolated global shared mutable state
   |                |- note: convert 'module' to a 'let' constant to make 'Sendable' shared state immutable
   |                |- note: annotate 'module' with '@MainActor' if property should only be accessed from the main actor
   |                `- note: disable concurrency-safety checks if accesses are protected by an external synchronization mechanism
```

This change makes sure it can build


cgrindel:
@maxwellE How do we configure a build to happen in Swift 6 mode? We should add that setting to our `resources_example`.

maxwellE:
> @maxwellE How do we configure a build to happen in Swift 6 mode? We should add that setting to our `resources_example`.

Is that blocked by having GithubActions agents to run on 16.2?

cgrindel:
They are all running on MacOS 15 with Xcode 16.2 as of #1533.

luispadron (approved):

cgrindel (approved):

cgrindel:
@mergifyio queue

PR #1566, opened 2025-04-10
fix: precompiled resource bundle warnings

sebastianv1:
For SPM modules that use precompiled resource bundles (i.e `.bundle` files), Bazel throws the warning
```
WARNING: input 'Sources/Down/Resources/DownView (macOS).bundle' is a directory; dependency checking of directories is unsound
```

Instead of just referencing the `.bundle` path directly we should wrap it in a `apple_bundle_import` for the prebuilt bundle and then add it into the `resources` attr.

### DIFF
--- a/.github/actions/tidy_and_test/action.yml
+++ b/.github/actions/tidy_and_test/action.yml
@@ -3,6 +3,11 @@ name: Tidy and Test
 runs:
   using: composite
   steps:
+    # DEBUG BEGIN
+    - name: Setup tmate session
+      uses: mxschmitt/action-tmate@v3
+    # DEBUG END
+
     - name: Ensure everything is tidy
       shell: bash
       run: |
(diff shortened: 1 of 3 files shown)


jpsim (approved):
Thanks!

cgrindel:
@sebastianv1 It looks like one of the examples is failing with this PR. The `tidy_and_test_matrix` can be fixed by running `bazel run //:tidy` and checking in the updated files.

sebastianv1:
@cgrindel Hmm i ran `bazel run //:tidy` and there aren't any changes locally for me? 

And I ran `//examples:firebase_example_test_bazel_.bazelversion` off `main` and that seems to fail as well for me; looks like we need to bump `rules_xcodeproj`. I can do that to unblock this PR as well everywhere.

cgrindel:
@mergifyio queue

cgrindel:
@mergifyio queue

cgrindel:
@mergifyio queue

cgrindel:
@mergifyio queue

cgrindel:
@mergifyio queue

Issue #1279, opened 2024-10-05
Support `embrace-io/embrace-apple-sdk`.


cgrindel:
Related [Slack thread](https://bazelbuild.slack.com/archives/C05NZSHE7FU/p1728004532149779).

There are two issues:
1. The [EmbraceObjCUtilsInternal](https://github.com/embrace-io/embrace-apple-sdk/tree/main/Sources/EmbraceObjCUtilsInternal) module has two module map files. The existing `rules_swift_package_manager` code treats this as an error. The fix is to warn the client and just use the first one that is found.
2. The [GRDB.swift package](https://github.com/groue/GRDB.swift/blob/dc03b8a29e2f5d51bd41e0971bba9a9975a29934/Package.swift#L46-L48), a transitive dependency, defines a `systemLibrary` target. `rules_swift_package_manager` does not yet support these target types.

adincebic:
@cgrindel what does it take to support packages with system library targets? Is the change complex?

cgrindel:
> what does it take to support packages with system library targets? Is the change complex?

Probably not. I have not dug into it. I would look at how SPM builds targets linked to these libraries. There is a flag which tells SPM to output its build commands. Then, it is a matter of generating the appropriate targets [here](https://github.com/cgrindel/rules_swift_package_manager/blob/107b2218db57581891db2fdb412e4ebd026c43ce/swiftpkg/internal/swiftpkg_build_files.bzl#L483-L485).

adincebic:
I'll have some time next weekend to attend this issue. Thank you both for references.

lucasmarcal-faire:
SystemLibrary Support Draft PR: #1323

luispadron:
Duplicate of #248, thanks for working on this @lucasmarcal-faire!

adincebic:
@cgrindel @0xLucasMarcal @luispadron 

There is a couple of developments here:
* Support for system libraries is no longer needed for Embrace SDK as they dropped GRDB.
* Embrace also no longer have multiple modulemap files.

However there is three outstanding issues:
* Embrace depends on [KSCrash](https://github.com/kstenerud/KSCrash) which requires [this PR](https://github.com/cgrindel/rules_swift_package_manager/pull/1594) to be merged.
[...]

PR #1594, opened 2025-05-02
fix: Always create cc_library targets if C/C++ files are present

jflan-dd:
I ran into an issue were we couldn't build the [InjectionNext](https://github.com/johnno1962/InjectionNext) package because there was a C file that wasn't being included in the `srcs` list of any generated target.

This was because most of the code is in Objective-C++, which creates a target that doesn't include C source files.

This change always creates a dedicated `cc_library` target for any C/C++ code in the package.


adincebic:
Thos PR helps in supporting KSCrash as well, see [this issue](https://github.com/cgrindel/rules_swift_package_manager/issues/1279)

cgrindel (approved):

PR #1621, opened 2025-05-10
fix: `swift_package_tool` so that it works on Linux and MacOS

### LINKED ISSUE #1591: `//:update_swift_packages_for_modified` does not appear to be working.
The Renovate PRs with modified `Package.swift` files do not have corresponding `Package.resolved` files.

cgrindel:
The Renovate runs have been failing when trying to run `//:update_swift_packages` on modified workspaces. The issue is that the Swift worker processes behave differently on MacOS vs Linux. On MacOS, the arguments are essentially passed through to `xcrun`. On Linux, it is more of a barebones process runner.

Previously, we expected the worker to find `swift` automatically. This does not work on Linux. This PR adds logic to find `swift` by attempting to ask the worker. If that fails, fall back to looking in the `PATH`.

I also upgraded Swift to 6.1 on Ubuntu.

Closes #1591.

### DIFF
--- a/.github/actions/set_up_ubuntu/action.yml
+++ b/.github/actions/set_up_ubuntu/action.yml
@@ -10,7 +10,7 @@ inputs:
     required: true
   swift_release_tag:
     description: The Swift release tag.
-    default: "swift-6.0.3-RELEASE"
+    default: "swift-6.1.0-RELEASE"
 
 runs:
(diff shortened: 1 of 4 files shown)


jpsim (approved):

cgrindel:
@mergifyio queue

PR #1625, opened 2025-05-12
fix: Define DEBUG for Clang targets in dbg

jflan-dd:
SPM conditionally defines `DEBUG=1` for Clang targets. We should too.

```swift
    /// A list of compilation conditions to enable for conditional compilation expressions.
    private var activeCompilationConditions: [String] {
        var compilationConditions = ["-DSWIFT_PACKAGE=1"]

        switch buildParameters.configuration {
        case .debug:
            compilationConditions += ["-DDEBUG=1"]
        case .release:
            break
        }

        return compilationConditions
    }
```
https://github.com/swiftlang/swift-package-manager/blob/2b0505e24798468102e9ec1937bb485346c59695/Sources/Build/BuildDescription/ClangModuleBuildDescription.swift#L448-L459

This was causing a linker issue in debug for InjectionNext which relies on DEBUG to conditionally include/exclude chunks of code in a C file with SPM
https://github.com/johnno1962/DLKit/blob/8619936ecb668618f95fe1507325ac6ee9710d5a/Sources/DLKitC/DLKitC.c#L14

### DIFF
--- a/swiftpkg/internal/swiftpkg_build_files.bzl
+++ b/swiftpkg/internal/swiftpkg_build_files.bzl
@@ -12,6 +12,10 @@ load(":pkginfo_target_deps.bzl", "pkginfo_target_deps")
 load(":pkginfo_targets.bzl", "pkginfo_targets")
 load(":pkginfos.bzl", "build_setting_kinds", "module_types", "pkginfos", "target_types")
 load(":starlark_codegen.bzl", scg = "starlark_codegen")
+load(
+    "//config_settings/bazel/compilation_mode:compilation_modes.bzl",
+    bazel_compilation_modes = "compilation_modes"
+)
 
 # MARK: - Target Entry Point
 
@@ -296,7 +300,7 @@ def _clang_target_build_file(repository_ctx, pkg_ctx, target):
     copts.append(
         bzl_selects.new(
             value = "-DDEBUG=1",
-            condition = "@rules_swift_package_manager//config_settings/bazel/compilation_mode:dbg",
+            condition = bazel_compilation_modes.label(bazel_compilation_modes.debug),
         ),
     )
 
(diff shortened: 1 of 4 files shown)


swiftpkg/internal/swiftpkg_build_files.bzl
```diff
@@ -292,6 +292,14 @@ def _clang_target_build_file(repository_ctx, pkg_ctx, target):
         "-fmodule-name={}".format(target.c99name),
     ]
 
+    # SPM defines DEBUG=1 for clang targets when building for debug
+    copts.append(
+        bzl_selects.new(
+            value = "-DDEBUG=1",
+            condition = "@rules_swift_package_manager//config_settings/bazel/compilation_mode:dbg",
```
jflan-dd, line 299: Is there a more idiomatic way to express this?

swiftpkg/internal/swiftpkg_build_files.bzl
```diff
@@ -296,7 +300,7 @@ def _clang_target_build_file(repository_ctx, pkg_ctx, target):
     copts.append(
         bzl_selects.new(
             value = "-DDEBUG=1",
-            condition = "@rules_swift_package_manager//config_settings/bazel/compilation_mode:dbg",
+            condition = bazel_compilation_modes.label(bazel_compilation_modes.debug),
```
cgrindel, line 303: I am being dense. How does this fix the issue? Can you explain why this is different from what we currently have checked into `main`?

jflan-dd: The string to `bazel_compilation_modes.label(...)` doesn't change anything, I was just trying to match the behavior of the SPM `configurations.label` to avoid the long string literal at the call site.

The entire `copts.append(bzl_select.new(value = "-DDEBUG=1`...)` block is new in this PR, which fixes the issue by defining `DEBUG` for clang targets when in `dbg` mode.

cgrindel:
The test/tidy errors appear to be due to `swiftpkg_build_files` now had a dependency on `//config_settings/bazel/compilation_mode:compilation_modes`.

cgrindel (approved):

PR #1636, opened 2025-05-18
feat: Support `env` and `env_inherit` attrs in `swift_deps` module extension

luispadron:
Adds support for setting `env` explicitly or inherting them via `env_inherit`. This is required for some packages as some of them use `ProcessInfo.processInfo.environment` to alter behavior. This env is passed down to the relevant repository rules.

Also added a test case via `interesting_deps` where if this is not working the `Package.swift` provides invalid dependencies that should fail the tests. With this change the tests pass.


cgrindel (commented):
Overall, LGTM. One comment regarding tabs/spaces. I will approve when I CI go green.

cgrindel (approved):

swiftpkg/internal/swift_package_tool_runner_template.sh
```diff
@@ -76,6 +76,14 @@ swift_executable="$(
 		)
 )"
 
+# Set the environment variables.
+env=(%(env)s)
+if [ -n "${env:-}" ]; then
+	for env_var in "${env[@]}"; do
+		export "$env_var"
+	done
+fi
```
luispadron, line 85:
```suggestion
if [ -n "${env:-}" ]; then
    for env_var in "${env[@]}"; do
        export "$env_var"
    done
fi
```

Issue #1547, opened 2025-03-25
Support multiple Package.swift file


saeed-rz:
Hi

I have a project with this structure.
```
Root
Source
Package
  Binaries
    Package.swift
  Services
    Package.swift
    Package.Resolved
  Features
    Package.swift
    Package.Resolved
  Shared
    Package.swift
```

Is there a way to support multiple Package.resovled?
I have this

```
swift_deps.from_package(
    declare_swift_deps_info = True,
    resolved = "//Packages:Features/Package.resolved",
    swift = "//Packages:Features/Package.swift",
)
```

but what should I do for other packages?

cgrindel:
It depends what you are hoping to accomplish. What is the purpose of having multiple `Package.swift` files? Do you publish multiple Swift packages?

saeed-rz:
Yes I'm moving from Xcode build to Bazel,
I was trying to keep the project architecture as it is, because I think it will take time to migrate everything.
hmm, I see, then I need to do more investigation. Do you have any examples or documentation?

saeed-rz:
Thank you
I will try the Option 2 first, but in the end I think I have to do it incremenrally

brentleyjones:
Btw, we also need to support multiple `Package.swift`. In particular, we want to scope things like how **rules_python** does it. We need to have tools and apps use different versions of dependencies.

cgrindel:
> In particular, we want to scope things like how rules_python does it. We need to have tools and apps use different versions of dependencies.

😢 

Supporting multiple versions of a dep will require some thought. My initial thoughts:

- Updating the extension API shouldn't be too big of a deal, given that bzlmod tag classes can be called multiple times already.
- We would need to add a setting/mode to `rules_swift_package_manager` that would generate version-specific repo names. We could detect the need for the setting if multiple `Package.swift`/`Package.resolved` files were specified.
- Not important for `rules_swift_package_manager`, but we might want to update `swift_gazelle_plugin` to handle this gracefully.

cgrindel:
@saeed-rz BTW, if you do not need to support different versions of your dependencies, you should be able to proceed with my previous advice. 

Best practices are not to introduce multiple versions of dependencies. However, technical and business considerations can trump are hopes and dreams.

brentleyjones:
> that would generate version-specific repo names

I don't think we need that. I think we can just create "hubs" for each `Package.swift`, and give it name name which ends up being the prefix instead of `swiftpkg_`.

[...]

PR #1696, opened 2025-06-23
feat: add ability to cache json responses in-repo

brentleyjones:
Besides a performance improvement, it allows cross-platform builds to work (e.g. Linux host building Apple code).

### DIFF
--- a/swiftpkg/internal/pkginfos.bzl
+++ b/swiftpkg/internal/pkginfos.bzl
@@ -28,6 +28,7 @@ def _get_dump_manifest(
         env = {},
         working_directory = "",
         debug_path = None,
+        cache_path = None,
         registries_directory = None,
         replace_scm_with_registry = False):
     """Returns a dict representing the package dump for an SPM package.
@@ -52,6 +53,10 @@ def _get_dump_manifest(
         debug_path = str(repository_ctx.path("."))
     debug_json_path = paths.join(debug_path, "dump.json")
 
+    cached_json_path = None
+    if cache_path:
+        cached_json_path = paths.join(cache_path, "dump.json")
+
     args = ["swift", "package"]
 
     if registries_directory:
@@ -66,6 +71,7 @@ def _get_dump_manifest(
         env = env,
         working_directory = working_directory,
         debug_json_path = debug_json_path,
+        cached_json_path = cached_json_path,
     )
 
 def _get_desc_manifest(
@@ -73,6 +79,7 @@ def _get_desc_manifest(
         env = {},
         working_directory = "",
         debug_path = None,
+        cache_path = None,
         registries_directory = None,
         replace_scm_with_registry = False):
     """Returns a dict representing the package description for an SPM package.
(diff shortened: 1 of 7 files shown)


jpsim:
Out of curiosity, can you please share more about the performance implications? What is improved and by how much?

brentleyjones:
That's actually the least important part. Whatever time is taken up by `swift package dump-package` and `swift package description` is now cached. But soon I expect we can use the repo cache that was introduced in Bazel 8.3.0 to save on the common case time there.

The important part is being able to force packages to be treated the same regardless of host OS. If you cache the packages on macOS, but build on Linux, they will now use the macOS resolutions. Without this you can run into build issues, like resource bundles missing (as they aren't supported on Linux).

This doesn't make things perfect, since you might want to resolve things for Linux as well on Linux. But I think #1547 and possibly extending this caching itself to support multiple OSes would be the solution there.

jpsim:
Thanks for the details! I don't use RBE for any of my projects, so I'm not directly impacted by the cross-platform caching aspect, so take this with a grain of salt, but wouldn't it be possible for some projects to actually prefer having the platform-specific implementations of `swift package dump-package` and `swift package description`? I can imagine some cross-platform Swift projects may want to have different dependencies or package manifest settings across macOS and Linux, and for those projects they actively wouldn't want to cache the results across macOS/Linux. It's probably the minority case though, so if there's a way to opt out of these files being cached across host OSes, that might be ok.

I'm also happy if this is something that we just handle if the need arises for it in the future.

So I'm ok with this change landing and will gladly take the perf improvements to swift package commands 👍

swiftpkg/internal/swift_package.bzl
```diff
@@ -197,7 +206,10 @@ _ALL_ATTRS = dicts.add(
     _GIT_ATTRS,
     repo_rules.env_attrs,
     repo_rules.swift_attrs,
-    {"version": attr.string(doc = "The resolved version of the package.")},
+    {
+        "cached_json_directory": attr.string(),
```
cgrindel, line 210: We should add a doc string.

swiftpkg/internal/repository_utils.bzl
```diff
@@ -78,17 +78,33 @@ def _parsed_json_from_spm_command(
...
+
     if debug_json_path:
         if not paths.is_absolute(debug_json_path):
             debug_json_path = paths.join(working_directory, debug_json_path)
         repository_ctx.file(debug_json_path, content = json_str, executable = False)
+    if cached_json_path:
+        if debug_json_path:
+            tmp_path = debug_json_path
+        else:
+            tmp_path = paths.join(str(repository_ctx.path(".")), paths.basename(cached_json_path))
```
cgrindel, line 103: Line length.

swiftpkg/internal/pkginfos.bzl
```diff
@@ -1648,7 +1664,9 @@ def _new_resource_rule_process(localization = None):
 
 def _new_resource_from_desc_map(desc_map, pkg_path):
     path = desc_map["path"]
-    if paths.is_absolute(path):
+    if path.startswith("./"):
+        path = path[2:]
```
cgrindel, line 1668: Can this be `path.removeprefix("./")`?

cgrindel (commented):
Cool idea. Should we add a utility to help a client cache their JSON?

PR #1702, opened 2025-06-25
fix: support implicit resource localization

jflan-dd:
Adding a dependency on https://github.com/mapbox/mapbox-navigation-ios caused the following error:

```
ERROR: .../rules_swift_package_manager++swift_deps+swiftpkg_mapbox_navigation_ios/BUILD.bazel:55:34: Label '@@rules_swift_package_manager++swift_deps+swiftpkg_mapbox_navigation_ios//:Sources/MapboxNavigation/Resources/Base.lproj/Navigation.storyboard' is duplicated in the 'resources' attribute of rule 'MapboxNavigation.rspm_resource_bundle'
```

This is because [this file](https://github.com/mapbox/mapbox-navigation-ios/tree/v2.18.4/Sources/MapboxNavigation/Resources/Base.lproj) was being added both from `desc_map` and as a discovered resource. They aren't being deduplicated in `resources_set` because the `desc_map` instance has localization `"Base"`, but the discovered resource had `None` for its localization.

This PR is intended to match the behavior of [this code](https://github.com/swiftlang/swift-package-manager/blob/2f03e7a16e2c4d08a3a16f16cca627ff53b70271/Sources/PackageLoading/TargetSourcesBuilder.swift#L312-L318) in SPM.

Alternatively we could deduplicate the labels themselves closer to where the build file is constructed if there's no need to track the localizations.


swiftpkg/internal/pkginfos.bzl
```diff
@@ -1680,10 +1680,22 @@ def _new_resource_from_discovered_resource(path):
...
 
+def _localization_from(path):
+    # Extract the implict localization from the path if its parent folder ends in ".lproj"
+    # https://github.com/swiftlang/swift-package-manager/blob/2f03e7a16e2c4d08a3a16f16cca627ff53b70271/Sources/PackageLoading/TargetSourcesBuilder.swift#L312-L318
+    parent_dir = paths.basename(paths.dirname(path))
+
+    (localization, extension) = paths.split_extension(parent_dir)
+
+    if extension == ".lproj":
+        return localization
```
cgrindel, line 1697: What should this function return if it does not have the `.lproj` extension? It seems like we are missing a return value for that case.

jflan-dd: It'd return `None`, which is what was being passed to `_new_resource_rule_process` before this change. I'll update it to make the `return None` explicit

Issue #1715, opened 2025-07-01
fix: don't rely on legacy `config_setting` visibility


MathieuTricoire:
Users who set the `--incompatible_config_setting_private_default_visibility` Bazel flag may encounter visibility errors when building generated targets that depend on `//config_settings/bazel/compilation_mode:....`

For example, this error occurred in one of my project:
```
ERROR: .../external/rules_swift_package_manager++swift_deps+swiftpkg_gtm_session_fetcher/BUILD.bazel:87:13: in objc_library rule @@rules_swift_package_manager++swift_deps+swiftpkg_gtm_session_fetcher//:GTMSessionFetcherCore.rspm_objc: Visibility error:
target '@@rules_swift_package_manager+//config_settings/bazel/compilation_mode:dbg' is not visible from
target '@@rules_swift_package_manager++swift_deps+swiftpkg_gtm_session_fetcher//:GTMSessionFetcherCore.rspm_objc'
Recommendation: modify the visibility declaration if you think the dependency is legitimate. For more info see https://bazel.build/concepts/visibility
```

Historically, Bazel has not enforced visibility for `config_setting` targets, as noted in the Bazel docs, we should avoid relying on this legacy behavior: https://bazel.build/concepts/visibility#config-setting-visibility

PR #1716, opened 2025-07-01
fix: set default visibility to public

MathieuTricoire:
I added `package(default_visibility = ["//visibility:public"])` to the `compilation_mode` package so its behavior matches the other `//config_settings/...` packages that already declare a public default.

Follow-up to discuss: should we add `--incompatible_config_setting_private_default_visibility` or `--incompatible_enforce_config_setting_visibility` to `.bazelrc`, I'm not sure if this could have unwanted effects.

fixes: #1715

### DIFF
--- a/config_settings/bazel/compilation_mode/BUILD.bazel
+++ b/config_settings/bazel/compilation_mode/BUILD.bazel
@@ -2,14 +2,13 @@ load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 load("@cgrindel_bazel_starlib//bzlformat:defs.bzl", "bzlformat_pkg")
 load(":compilation_modes.bzl", "compilation_modes")
 
-package(default_visibility = ["//visibility:public"])
-
 bzlformat_pkg(name = "bzlformat")
 
 [
     config_setting(
         name = comp_mode,
         values = {"compilation_mode": comp_mode},
+        visibility = ["//visibility:public"],
     )
     for comp_mode in compilation_modes.all_values
 ]
--- a/config_settings/bazel/compilation_mode/BUILD.bazel
+++ b/config_settings/bazel/compilation_mode/BUILD.bazel
@@ -2,6 +2,8 @@ load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 load("@cgrindel_bazel_starlib//bzlformat:defs.bzl", "bzlformat_pkg")
 load(":compilation_modes.bzl", "compilation_modes")
 
+package(default_visibility = ["//visibility:public"])
+
 bzlformat_pkg(name = "bzlformat")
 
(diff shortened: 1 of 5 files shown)


config_settings/bazel/compilation_mode/BUILD.bazel
```diff
@@ -2,6 +2,8 @@ load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 load("@cgrindel_bazel_starlib//bzlformat:defs.bzl", "bzlformat_pkg")
 load(":compilation_modes.bzl", "compilation_modes")
 
+package(default_visibility = ["//visibility:public"])
+
```
cgrindel, line 6: Instead of setting this on the package, let's set the `visibility` attribute on the `config_setting`.

cgrindel (commented):
Since you were kind enough to submit a PR, would you mind updating the other `config_setting` and `selects.config_setting_group`?

https://github.com/cgrindel/rules_swift_package_manager/blob/221b262c8a2fd4d4bf686c6e85a7a13db9a262da/config_settings/spm/platform/BUILD.bazel#L17
https://github.com/cgrindel/rules_swift_package_manager/blob/221b262c8a2fd4d4bf686c6e85a7a13db9a262da/config_settings/spm/platform/BUILD.bazel#L31
https://github.com/cgrindel/rules_swift_package_manager/blob/221b262c8a2fd4d4bf686c6e85a7a13db9a262da/config_settings/spm/platform_configuration/BUILD.bazel#L11
https://github.com/cgrindel/rules_swift_package_manager/blob/221b262c8a2fd4d4bf686c6e85a7a13db9a262da/config_settings/spm/configuration/BUILD.bazel#L18
https://github.com/cgrindel/rules_swift_package_manager/blob/221b262c8a2fd4d4bf686c6e85a7a13db9a262da/config_settings/bazel/apple_platform_type/BUILD.bazel#L10
https://github.com/cgrindel/rules_swift_package_manager/blob/221b262c8a2fd4d4bf686c6e85a7a13db9a262da/config_settings/bazel/compilation_mode/BUILD.bazel#L8

config_settings/bazel/compilation_mode/BUILD.bazel
MathieuTricoire: Done, I had to set `visibility = ["//visibility:public"]` to associated `bzl_library` rules, let me know if it would be better to set a different level of visibility.

brentleyjones (approved):

cgrindel:
Released in https://github.com/cgrindel/rules_swift_package_manager/releases/tag/v1.3.0.
It will be available from the BCR when https://github.com/bazelbuild/bazel-central-registry/pull/5064 merges.

Issue #1743, opened 2025-07-17
Not applying Swift 6 language mode when defined at the package level


jflan-dd:
Packages that define a `swift-tools-version` of 6.0+ automatically use Swift 6 language mode unless a target explicitly opts back to Swift 5 mode.

https://www.swift.org/migration/documentation/swift-6-concurrency-migration-guide/swift6mode/#Package-manifest

It looks like RSPM has support for individual targets opting into Swift 6 mode, but not at the package level.
https://github.com/cgrindel/rules_swift_package_manager/blob/0eef5048be38f776f1866f354fe150ef473aea76/swiftpkg/internal/swiftpkg_build_files.bzl#L144-L151

`swift package dump-package` contains this array when using Swift 6 tools:
```
  "swiftLanguageVersions" : [
    "6"
  ],
```

and `swift package describe` contains this array:
```
  "swift_languages_versions" : [
    "6"
  ],
```

PR #1770, opened 2025-08-08
fix: honor package level language mode configuration

jflan-dd:
Ensure targets are built with the correct Swift language mode based on package level information.

Fixes https://github.com/cgrindel/rules_swift_package_manager/issues/1743


examples/language_modes_example/MODULE.bazel
```diff
@@ -0,0 +1,50 @@
...
+swift_deps = use_extension(
+    "@rules_swift_package_manager//:extensions.bzl",
+    "swift_deps",
+)
+swift_deps.from_package(
+    declare_swift_deps_info = True,
+    resolved = "//:Package.resolved",
+    swift = "//:Package.swift",
+)
+use_repo(swift_deps, "swift_deps_info", "swift_package", "swiftpkg_swift_5_package", "swiftpkg_swift_5_package_v6", "swiftpkg_swift_6_package", "swiftpkg_swift_log")
```
cgrindel, line 50: Please wrap this line to not exceed 80 characters.

examples/language_modes_example/README.md
cgrindel, line 1: Please wrap markdown lines to not exceed 100 characters.

cgrindel (commented):
I added a couple of formatting comments. Just let us know when tests are green and we can take another look.

jflan-dd:
The remaining failing test is failing with:
```
error: failed retrieving apple.swift-system version 1.4.0 manifest from https://fake-spm-registry.deno.dev/: server error 500: Internal Server Error
```
which I don't think because of this PR

jflan-dd:
I think we should support pinning the language mode down to Swift 5 and only use `-swift-version X` to set the package language mode by removing the use of `swift.enable_v6`

The v6 feature potentially has other effects if the user isn't using the Swift 6+ compiler that package authors wouldn't expect. 

```
# If this feature is enabled, the toolchain should enable the features that are
# available in Swift 6 language mode. If the toolchain supports
# `-swift-version 6`, it will do so using that flag. If it is older, it will
# enable the set of upcoming features that will be on by default in Swift 6,
# allowing users to prepare their code base by opting in to the full set of
# Swift 6 features even before switching to a Swift 6 compiler.
```
https://github.com/bazelbuild/rules_swift/blob/ccebe4dc4ac16ea45ef111bee33f31036027ac08/swift/internal/feature_names.bzl#L33-L39

I believe if the package used `swiftLanguageVersions: [.version("6"), .v5]` as documented in the [the migration guide](https://www.swift.org/migration/documentation/swift-6-concurrency-migration-guide/swift6mode/#Package-manifest) while using the Swift 5 compiler then we'd end up setting the `swift.enable_v6` feature which would enable a bunch of language features that the package author didn't expect to be enabled for the Swift 5 compiler.

swiftpkg/internal/semver.bzl
cgrindel, line 1: Can we use [Skylib's versions module](https://github.com/bazelbuild/bazel-skylib/blob/main/lib/versions.bzl)?

cgrindel: Let's go with your version. Thanks!

cgrindel (approved):

cgrindel:
@mergifyio queue

cgrindel:
@mergifyio queue

PR #1792, opened 2025-08-27
fix: work around SPM executable binary target bug

jflan-dd:
Ignore errors from `swift package` commands caused by executable products depending on binary targets

Work around https://github.com/cgrindel/rules_swift_package_manager/pull/1781#issuecomment-3217081119

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/bazelbuild/buildtools v0.0.0-20240918101019-be1c24cc9a44
 	github.com/creasty/defaults v1.8.0
 	github.com/deckarep/golang-set/v2 v2.8.0
-	github.com/spf13/cobra v1.10.0
+	github.com/spf13/cobra v1.10.1
 	github.com/stretchr/testify v1.11.1
 	golang.org/x/exp v0.0.0-20250819193227-8b4c13bb791b
 	golang.org/x/text v0.28.0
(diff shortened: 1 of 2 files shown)


swiftpkg/internal/repository_utils.bzl
```diff
@@ -56,7 +56,7 @@ def _execute_spm_command(
         environment = exec_env,
         working_directory = working_directory,
     )
-    if exec_result.return_code != 0:
+    if exec_result.return_code != 0 and not _can_ignore_spm_error(exec_result):
```
cgrindel, line 59: Are we sure this will work? We still need the JSON. I believe it will not exist if the command fails.

jflan-dd: It outputs valid JSON to stdout in addition to writing the error to stderr. The "problematic" product is missing, but I doubt for this use case anyone intending to use the executable product.
image

cgrindel (approved):

cgrindel (changes requested):
It looks like tidy is failing. I saw at least one issue with the doc string for the new function. Once things are green, we should be good to go.

jflan-dd:
@cgrindel I updated the doc comment. `bazel run //:tidy` didn't catch it and I didn't realize that `bazel test //...` would test only the internal rule tests rather than also running the various example tests. I'll add that to my workflow when contributing here.

cgrindel (approved):

cgrindel:
> @cgrindel I updated the doc comment. `bazel run //:tidy` didn't catch it and I didn't realize that `bazel test //...` would test only the internal rule tests rather than also running the various example tests. I'll add that to my workflow when contributing here.

If you want to run the examples locally, there are `//:smoke_integration_tests` and `//:all_integration_tests`. However, I tend to create a draft PR and let GitHub run them in parallel.

PR #1880, opened 2025-10-19
fix: force symlink dereference when using file command

ghugues:
On macOS, the `file` command dereferences symbolic links by default (`--dereference` implied). On Linux it's the opposite (`--no-dereference` implied).

We want the dereference behavior on both, otherwise `file` outputs `symbolic link to Versions/Current/MyFrameworkBinary` on Linux when checking the file type of an iOS framework (we then fallback to the `otool` command which fails).


cgrindel (changes requested):
Thank you for the contribution! It looks like a test is failing. Can you take a look?

PR #1967, opened 2026-01-03
feat: add GRDB Linux example with system library and symlink loop fixes

jpsim:
## Summary

- Add GRDB.swift example that works on both macOS and Linux
- Fix filesystem loop detection for GNU `find` on Linux (GRDB has symlink loops in test directories)
- Fix system library linking by extracting `link` declarations from modulemaps
- Fix system library modulemap discovery for targets using `.systemLibrary`

## Details

This PR adds a new GRDB example and fixes several bugs discovered while making it work on Linux:

1. **Filesystem Loop Detection** (`repository_files.bzl`): GNU `find` on Linux fails when it encounters symlink loops (like in GRDB's test directory). The fix detects this specific error and continues with partial results.

2. **System Library Linking** (`swiftpkg_build_files.bzl`): System library targets can declare link dependencies via `module.modulemap` files (e.g., `link "sqlite3"`). These were being ignored, causing linker errors. The fix parses modulemaps and adds `-l` linkopts.

3. **System Library Modulemap Discovery** (`pkginfos.bzl`): System library targets weren't populating `clang_src_info`, so the modulemap path was always `None`. The fix adds special handling for system library targets.

4. **GRDB Snapshot Compatibility**: Uses `configure_package` with `patch_cmds` to add `GRDBCUSTOMSQLITE` define, disabling GRDB's snapshot code path which requires SQLite functions not available on Linux.

## Test plan

- [x] Linux (Docker): `cd examples/grdb_example && ./docker_test.sh`
- [x] macOS: `cd examples/grdb_example && bazel run //:tidy && bazel build //... && bazel run //Sources/GRDBExample`


cgrindel (approved):
LGTM. Thank you!

cgrindel:
@mergifyio queue

PR #2022, opened 2026-02-02
fix: exclude paths not working when not at root of repo

luispadron:
Not sure how this was working before (doesnt seem like it?) but for example, the current behavior was returning resource paths within `ios/Market/App` even though `ios/Market` is a target with `excludes: ["App"]`. The issue is it was checking if `ios/Market/App` starts with `App` in order to exclude but in this case it needs to be aware of the `path` for the target.


cgrindel:
@luispadron Oops. I forgot that I had it in the merge queue. Let me know if you want any changes. Sorry about that.

PR #2024, opened 2026-02-02
fix: ignore requires keyword in modulemap

luispadron:
Was failing to when parsing `requires` declarations: https://clang.llvm.org/docs/Modules.html#requires-declaration it seems like these can be skipped instead of us failing on them for valid modulemaps like: https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogProfiling/Mach/include/module.modulemap


cgrindel:
@mergifyio queue

cgrindel:
@mergifyio requeue

PR #2073, opened 2026-02-24
feat: add ability to cache json responses in-repo

adincebic:
Besides a performance improvement, it allows cross-platform builds to work (e.g. Linux host building Apple code).

Cherry picked from https://github.com/cgrindel/rules_swift_package_manager/pull/1696

### DIFF
--- a/examples/aws_sdk_example/Package.resolved
+++ b/examples/aws_sdk_example/Package.resolved
@@ -5,8 +5,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/swift-server/async-http-client.git",
       "state" : {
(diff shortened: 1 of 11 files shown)


adincebic:
> Overall, LGTM. Can we update one of the examples to demonstrate the use of this feature. We should update the example's README.md explaining that it demonstrates the use of the feature.

Done!

cgrindel (approved):
Thank you!

iMostfa:
@adincebic Hi, QQ, is it expected to need to delete the caches JSONs when i add a new package to SPM ? 
if i don't delete it, no any JSON get added for the new dependency in SPM, so bazel mod tidy removes the new module..etc

Issue #2137, opened 2026-03-22
Support dependent trait enablement with union semantics (SPM traits Phase 2)


cgrindel:
## Summary

Follow-up to #2136 (Phase 1: default trait support). This issue tracks support for **dependent trait enablement** — allowing Bazel users to enable specific non-default traits on their Swift package dependencies, with union semantics matching SPM's behavior.

## Background: How SPM Trait Enablement Works

[SE-0450: Package Traits](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0450-swiftpm-package-traits.md) defines the following mechanics for trait activation by dependents:

### Enabling traits on dependencies

A package can enable specific traits on its dependencies via the `traits:` parameter:

```swift
.package(url: "https://github.com/example/pkg", from: "1.0.0", traits: ["FeatureA", .defaults])
```

- When `traits:` is omitted, only the dependency's `default` trait (and its transitive `enabledTraits`) are active.
- When `traits:` is specified, `.defaults` must be explicitly included to retain default traits.
- See: [Activating traits of a dependency](https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/packagetraits/)

### Union semantics

When multiple packages depend on the same package but request different traits, SPM **unions** all requested traits. For example:

- Package A depends on C with `traits: [.defaults, "Foo"]`
- Package B depends on C with `traits: [.defaults, "Bar"]`
- C is built with traits: `default` + `Foo` + `Bar` (union of all requests)

This is analogous to how SPM handles version resolution — the resolved version satisfies all dependents.

### Bazel challenge

In Bazel, each external package is built exactly once (single configuration). There is no mechanism today to collect trait requests from multiple dependents and union them. This requires a design for plumbing trait selections through the module extension (`swift_deps`), likely involving:

- A new attribute on `swift_deps.from_package` or a dedicated rule for specifying trait overrides
- Logic in the module extension to collect and union trait requests across all dependents
- Passing the resolved trait set to the repository rule that builds the package

## References

- [SE-0450: Package Traits](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0450-swiftpm-package-traits.md)
- [Swift Package Manager: Package Traits Documentation](https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/packagetraits/)
- [Swift Forums: SE-0450 Review](https://forums.swift.org/t/se-0450-package-traits/75598)
- Parent issue: #2136

PR #2138, opened 2026-03-22
feat: add SPM trait support for default traits (SE-0450)

### LINKED ISSUE #2136: feat: support Swift Package Manager traits (SE-0450)
## Summary

`rules_swift_package_manager` does not support [Swift Package Manager traits](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0450-swiftpm-package-traits.md) (SE-0450, introduced in Swift 6.1). This causes build failures when a dependency uses traits to conditionally enable build settings or features.

## Problem

When processing `dump.json`, the build file generator ignores the `condition.traits` field on build settings. Settings that should only apply when a specific trait is enabled are applied unconditionally. This leads to:

1. **Incorrect compiler defines** being passed (e.g., `-DCOLLECTIONS_UNSTABLE_CONTAINERS_PREVIEW` from `swift-collections` 1.4.0 is applied even though the `UnstableContainersPreview` trait is not in the default enabled set)
2. **Build failures** when trait-gated code requires a newer Swift compiler or features not available in the build environment
3. **Missing automatic trait defines** — SPM automatically passes enabled trait names as `-D` flags to the compiler, but `rules_swift_package_manager` does not

## Reproduction

The `grpc_example` fails on macOS when `swift-collections` is resolved to 1.4.0 (see PR #2104). The `ContainersPreview` module fails to compile because `~Escapable`/`@_lifetime` features are gated behind the `UnstableContainersPreview` trait, but the define is applied unconditionally.

## How SPM Traits Work

- Packages declare traits in `Package.swift` with a `default` trait specifying which are enabled by default
- Dependents can enable specific traits via `traits:` parameter on `.package(...)`
- Build settings and target dependencies can be conditioned on traits via `condition.traits`
- **Enabled trait names are automatically passed as `-D` flags** to the Swift compiler
- In `dump.json`:
  - Top-level `traits` array defines all traits; the `"default"` trait's `enabledTraits` lists default-active traits
  - Build settings have `condition.traits` arrays
  - `desc.json` does **not** include trait information

### Phase 1: Respect trait conditions (minimum viable fix)

1. **Parse the top-level `traits` array** from `dump.json` to determine the default enabled traits
2. **Filter build settings** — skip settings whose `condition.traits` references traits not in the enabled set
3. **Pass enabled trait names as `-D` flags** to the Swift compiler for all targets in the package

### Phase 2: Full trait support (#2137)

4. Support dependents enabling traits on their dependencies (requires Bazel-side configuration mechanism)
5. Handle trait-conditioned target dependencies (filter `deps` based on enabled traits)
6. Support transitive trait enablement (`enabledTraits` on individual traits)

## Affected Code

- `swiftpkg/internal/pkginfos.bzl` — `_new_build_setting_condition_from_json` ignores `traits` field
- `swiftpkg/internal/pkginfos.bzl` — `_new_*_settings_from_dump_json_list` functions need trait filtering
- `tools/swift_deps_index/internal/spdump/target.go` — `TargetSetting.UnmarshalJSON` ignores `condition`
- `swiftpkg/internal/swiftpkg_build_files.bzl` — needs to add trait `-D` flags to copts

## Related

- PR #2104 (grpc_example CI failure triggered by swift-collections 1.4.0)
- [SE-0450: Package Traits](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0450-swiftpm-package-traits.md)

cgrindel:
## Summary

Closes #2136

- Parse traits from `dump.json` and compute enabled defaults with transitive resolution
- Filter build settings by trait conditions — settings conditioned on non-enabled traits are skipped
- Filter target dependencies by trait conditions
- Add `-D<TraitName>` compiler flags for each enabled trait (mirrors SPM behavior)
- Update `grpc_example` deps to reproduce and verify the fix (swift-collections 1.4.0)

## Background

SPM traits (SE-0450, Swift 6.1+) allow packages to declare optional features. The `default` trait
controls which traits are active by default. Build settings and target dependencies can be
conditioned on traits.

The bug: `pkginfos.bzl` ignored `condition.traits` on build settings, treating trait-conditioned
settings as unconditional. For `swift-collections` 1.4.0, this caused
`-DCOLLECTIONS_UNSTABLE_CONTAINERS_PREVIEW` to be applied unconditionally, activating experimental
code in `ContainersPreview` that failed to compile.

## Phase 2

Dependent trait enablement (allowing consumers to enable non-default traits) is tracked in #2137.

## Test plan

- [x] Unit tests for trait parsing (default, transitive, cyclic, pre-6.1 packages)
- [x] Unit tests for build setting filtering by trait conditions
- [x] Unit tests for target dependency filtering by trait conditions
- [x] Unit test for BUILD file generation with trait `-D` flags
- [x] `grpc_example` integration test passes with swift-collections 1.4.0
- [x] All 341 existing unit tests pass

### DIFF
--- a/swiftpkg/internal/repository_utils.bzl
+++ b/swiftpkg/internal/repository_utils.bzl
@@ -107,7 +107,8 @@ def _parsed_json_from_spm_command(
         arguments,
         env = env,
         working_directory = working_directory,
-    ).replace(working_directory + "/", "./")
+    )
+    json_str = _replace_working_directory(json_str, working_directory)
 
     if debug_json_path:
         if not paths.is_absolute(debug_json_path):
@@ -163,10 +164,34 @@ def _struct_to_kwargs(*, struct, keys):
             kwargs[k] = v
     return kwargs
 
+def _replace_working_directory(json_str, working_directory):
+    """Replace the working directory prefix in a JSON string.
+
+    Uses a path-prefix-safe replacement by appending a trailing
+    slash to ensure only full path-prefix matches are replaced.
+    Without the trailing slash, a working directory like
+    "/path/to/MyApp" would incorrectly match
+    "/path/to/MyAppFrameworks".
+
+    Args:
+        json_str: A `string` containing JSON output from an
+            SPM command.
+        working_directory: A `string` representing the working
+            directory path.
+
+    Returns:
+        A `string` with the working directory prefix replaced
+        by "./".
+    """
+    if not working_directory:
+        return json_str
+    return json_str.replace(working_directory + "/", "./")
+
 repository_utils = struct(
     exec_spm_command = _execute_spm_command,
     is_macos = _is_macos,
     package_name = _package_name,
     parsed_json_from_spm_command = _parsed_json_from_spm_command,
+    replace_working_directory = _replace_working_directory,
     struct_to_kwargs = _struct_to_kwargs,
 )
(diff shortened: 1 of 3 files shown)


cgrindel:
@mergifyio queue

Issue #1340, opened 2024-11-15
Feature request: update file permissions to enforce read-only files


luispadron:
The files from `swiftpkg` repositories should be marked read-only (locked) when creating the repositories. This helps ensure users do not accidentally modify a packages contents.

cgrindel:
Curious. What is the scenario where they are editing the files? Are folks digging into the `external` directory and modifying files?

luispadron:
I imagine it's when using rules_xcodeproj since that makes it a lil easier to view the external files / when viewing external sources and accidentally editing a file

cgrindel:
This is being worked by @adincebic

adincebic:
Completed in https://github.com/cgrindel/rules_swift_package_manager/pull/2143

PR #2143, opened 2026-03-28
feat: make fetched Swift package files read-only

adincebic:
Prevents accidental edits when using `rules_xcodeproj`.

Addresses: https://github.com/cgrindel/rules_swift_package_manager/issues/1340


cgrindel (commented):
This will be a behavioral change for all users. Should we put this behind a flag so that folks can opt in?

adincebic:
> > > This will be a behavioral change for all users. Should we put this behind a flag so that folks can opt in?
> 
> > 
> 
> > I would rather it being opt out but no strong opinion.
> 
> 
> 
> It is a new behavior. I think we need it to be opt-in. 

Idea was to match what SPM does. But I suppose it can be opt in

adincebic:
@cgrindel I picked this up again and added opt-in flag which defaults to False.

cgrindel (approved):
Thank you!

cgrindel:
@mergifyio queue

PR #2171, opened 2026-04-06
fix: respect DEVELOPER_DIR for Xcode version alignment in SPM commands

### LINKED ISSUE #2140: The wrong version of Swift is used when dumping packages
## Summary

Internally, `rules_swift_package_manager` uses `xcrun swift package` (or equivalent) to inspect Swift packages. This uses the Xcode version selected with `xcode-select`. Bazel normally doesn't work like this (or at least not always). The Xcode version can be selected with `--xcode_version_config` and be different from the one reported by `xcode-select -p`.

The result is that the Bazel build files generated from the Swift package dump can use the wrong package file (when multiple are provided for different versions of the tool) and be incompatible with the Xcode version that Bazel then uses. When these targets are built, it results in build failures.

Overall, Swift package authors expect that the Swift version used to resolve the package and the one used to compile the code is the same. When this is not guarantied, build failures can happen.

Take for example the [swift-collections](https://github.com/apple/swift-collections) package. If Xcode 16 is selected with `xcode-select` and used to construct the build files, but the project is then compiled with Xcode 26, compilation fails. The reason is that [LifetimeOverride.swift](https://github.com/apple/swift-collections/blob/1.4.1/Sources/InternalCollectionsUtilities/LifetimeOverride.swift) has a `#if compiler(>=6.2)`, and the code inside this check expects a number of Swift experimental features to be enabled (in this case Lifetimes), but if the package was resolved with an older version of the tools, then the feature is not enabled and compilation fails.

cgrindel:
## Summary

- Auto-inherit `DEVELOPER_DIR` from the environment in the `swift_deps` module extension so SPM
  commands use the same Xcode that Bazel will use at build time
- Propagate `env` to transitive local dependency inspection (previously missing)
- Declare `environ = ["DEVELOPER_DIR"]` on all repository rules and the module extension so Bazel
  invalidates cached repos when `DEVELOPER_DIR` changes

Closes #2140

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -3,6 +3,7 @@
 ## Table of Contents
 
 <!-- MARKDOWN TOC: BEGIN -->
+* [How do I ensure SPM resolution uses the correct Xcode version?](#how-do-i-ensure-spm-resolution-uses-the-correct-xcode-version)
 * [Does this replace [rules_spm]?](#does-this-replace-rules_spm)
 * [Where is the Swift gazelle plugin?](#where-is-the-swift-gazelle-plugin)
 * [After running `//:swift_update_pkgs`, I see a `.build` directory. What is it? Do I need it?](#after-running-swift_update_pkgs-i-see-a-build-directory-what-is-it-do-i-need-it)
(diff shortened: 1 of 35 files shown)


keith (approved):

ghugues (approved):

adincebic (approved):
LGTM. We added this change as a local patch and will report back if we hit any issues.

adincebic (changes requested):
After testing a bit I don't think this can land as is because the env var is changing depending on the machine. In practice this means that `MODULE.bazel.lock` file is different per machine. For us  it manifests on our Linux runners where we execute `bazel mod deps --lockfile_mode=error`.

cgrindel:
@adincebic Good catch. I am going to close this for now. I going to take another crack at #2140 after #2198 lands.

PR #2246, opened 2026-04-30
Remove .swift-version files before running spm

keith:
Swiftly always respects .swift-version files, which can lead to spm
commands failing if the repo's requested version isn't installed.
Ideally there would be an env var / flag to stop this upstream:
https://github.com/swiftlang/swiftly/issues/504


keith:
I don't know this repo well enough but could deleting a file here actually delete it in some local checkout of the repo?

adincebic (approved):

cgrindel:
@keith Claude and I created #2258 as an alternative for this PR. The primary difference is that we do the deletion once when the repo is set up. Does this address the issue as you understand it?

cgrindel:
Opted to merge #2258.

PR #2251, opened 2026-05-01
fix: normalize SwiftPM platform names

jpsim:
Lowercase PackageDescription spellings such as `iOS`, `macOS`, and
`visionOS` to RSPM's canonical platform names so config-setting
labels, supported-platform filtering, and parsed package platform
data stay consistent regardless of which SwiftPM JSON spelling
produced the name.

* Add `platforms.normalize` and route `label`, `is_supported`, and
  `supported` through it.
* Apply normalization in `pkginfos._new_platform` so parsed package
  platform names match the canonical spelling.
* Expose `visionos` on the `platforms` struct so callers do not
  need to spell it as a raw string.
* Add `spm_platforms_tests` covering normalization, label
  construction, and `supported` filtering.

PR #2252, opened 2026-05-01
feat: add minimum OS transition wrappers

jpsim:
Add internal `spm_minimum_os_target`, `spm_minimum_os_binary`, and `spm_minimum_os_test` rules. Each applies an incoming-edge transition that overrides the active Apple platform's minimum OS command-line option from per-platform wrapper attributes (`ios_minimum_os`, `macos_minimum_os`, etc.).

The wrappers forward `DefaultInfo`, `CcInfo`, `SwiftInfo`, `SwiftBinaryInfo`, `RunEnvironmentInfo`, `OutputGroupInfo`, `InstrumentedFilesInfo`, `testing.ExecutionInfo`, and `apple_common.Objc` so they can sit at a public generated label while the underlying implementation target builds under the package deployment target. The executable and test variants additionally symlink the dep's executable so the wrapper itself is runnable.

Centralize per-platform metadata (attr name, command-line option, and oldest-supported fallback) in a new `minimum_os_platforms` module and refactor `minimum_os_versions` to consume it.

Fail the build when an importer's minimum OS is lower than an imported target's minimum OS via a new `_import_error` check inside the transition; the message names the importing target and suggests raising its minimum OS to satisfy the imported one.

Cover the transition logic, the import validation, and the forwarded provider surface with unit and analysis tests under `swiftpkg/tests/minimum_os_transition/`.

Stacked on #2253 — the cumulative diff includes that PR's changes (which in turn includes #2251) until they merge; review focus here is the third commit.


cgrindel:
@jpsim Can you update this PR's upstream branch to point to the branch for #2253? It will make it easier to review.

jpsim:
I'll rebranch once the dependent PRs are merged. I can't target the other branches right now because I made these PRs from a fork.

cgrindel (approved):
LGTM. Thank your for all of the tests. We will need to rebase after #2253 has merged.

PR #2257, opened 2026-05-02
feat(swiftpkg): cache SPM dump/describe to avoid toolchain mismatch (GH-2140)

### LINKED ISSUE #2140: The wrong version of Swift is used when dumping packages

cgrindel:
## Summary

Fixes GH-2140. SPM inspection (`swift package dump-package` / `describe`) runs against whatever toolchain `xcrun` picks, while Bazel compiles against `--xcode_version_config`. When those disagree, `#if compiler(>=X.Y)` gets evaluated by one toolchain and compiled by another, and the generated BUILD files quietly describe a package you aren't actually building.

So: let people vendor SPM's `dump-package` and `describe` output into the repo and have the repo rules read that instead of shelling out to `xcrun` at fetch time. An auto-emitted `swift_info_test` checks the cached Swift version against the toolchain Bazel resolves, so a stale cache fails a test rather than silently generating the wrong BUILD files.

## What's here

- New `dump_manifests` / `desc_manifests` on `swift_deps`, and `cached_dump_manifest` / `cached_desc_manifest` on the `swift_package` family. Unset means today's `xcrun` behavior, so this is opt-in. `cached_json_directory` is deprecated in favor of them, and setting both is an error.
- `bazel run @swift_package//:cache` generates and refreshes the cache. It tolerates a missing `Package.resolved`, so first-time setup is one command instead of a separate `swift package update`.
- The cache is portable across machines: workspace paths are stored as `{{WORKSPACE_ROOT}}/<rel>` and expanded at fetch time. Squeezing out the last absolute paths took a couple of passes, since SPM canonicalizes symlinks in what it emits (macOS `/var` -> `/private/var` is the easy way to hit it), so we match both spellings of the prefix.
- Dropped `--swift_worker` from the cache tool; GH-2344 removed that mechanism upstream, so it resolves `swift` the way the resolve/update commands do.
- New `docs/swift_package_caching.md` covers setup, refresh, and drift; the README quickstart leads with the cache; `examples/pkg_manifest_minimal` exercises it end to end.

## Open question: cross-toolchain caching

A committed cache stamps one (OS, toolchain) pair, so a repo building on both macOS (Apple Swift) and Linux (open-source Swift) can't share one — `swift_info_test` fails on whichever side doesn't match. The auto-emitted test pins to the host OS so cross-platform CI skips rather than fails, but that papers over the limitation.

Directions I see: per-platform cache directories keyed by host triple; toolchain-specific `swift_info.json` next to a shared dump/desc, assuming the JSON rarely diverges; or call this a single-toolchain feature and document that mixed-toolchain repos stay on the legacy `xcrun` path. Would like feedback before picking one, especially from anyone building with both Apple and open-source Swift toolchains.


adincebic:
I plan to test this soon and provide feedback.

brentleyjones:
I tested this patch and we still have full absolute paths in the files.

PR #2275, opened 2026-05-07
feat: add API for manually extending deps on targets

keith:
Swift package manager doesn't currently support adding dependencies in
your Package.swift for `binaryTarget` targets. Previously this didn't
really matter as long as all the necessary binaries were all included in
the `deps` of the target that consumed them later. With explicit modules
the compilation is shifted from when it's used as a dependency to
`apple*framework_import` rule itself to create a PCM for the dependency.
The PCM compile needs all the correct deps. This allows users to
manually map the missing deps into the BUILD file.


keith:
example usage:

```bzl
swift_deps.configure_package(
    name = "swiftpkg_misnap_ios",
    target_deps = {
        "MiSnap": [
            ":MiSnapCamera",
            ":MiSnapCore",
            ":MiSnapScience",
        ],
        "MiSnapFacialCapture": [":MiSnapCore"],
        "MiSnapFacialCaptureUX": [
            ":MiSnapAssetManager",
            ":MiSnapCore",
            ":MiSnapFacialCapture",
        ],
        "MiSnapNFC": [":MiSnapCore"],
        "MiSnapNFCUX": [
            ":MiSnapAssetManager",
            ":MiSnapCore",
            ":MiSnapNFC",
        ],
        "MiSnapUX": [
            ":MiSnap",
            ":MiSnapAssetManager",
            ":MiSnapCamera",
            ":MiSnapCore",
            ":MiSnapScience",
        ],
        "MiSnapVoiceCapture": [":MiSnapCore"],
        "MiSnapVoiceCaptureUX": [
            ":MiSnapAssetManager",
            ":MiSnapCore",
            ":MiSnapVoiceCapture",
        ],
    },
)
```

cgrindel (approved):
Thanks! Is there an example that we can add that exercises this functionality? If we can identify one, it would be great to add it in a follow-up.

cgrindel:
@mergifyio queue

PR #2278, opened 2026-05-07
fix: put textual headers in modulemaps instead of directory

keith:
This is a different approach to 57b98a8987ea585b9d1d90466da564210844ad58
where instead of putting the path to the directory in the modulemap we
propagate all the headers but we include the non-importable ones as
textual headers. This solves an issue for precompiled modules where the
header directory was missing from building pcms for libraries.


swiftpkg/internal/generate_modulemap.bzl
```diff
@@ -31,43 +31,25 @@ def _generate_modulemap_impl(ctx):
...
-                umbrella_dir = "/".join(common_parts)
+    public_hdrs = []
+    textual_hdrs = []
+    for f in ctx.files.hdrs:
+        if f.path.endswith(".inl"):
+            textual_hdrs.append(f)
+        elif clang_files.is_hdr(f.path):
+            public_hdrs.append(f)
+        else:
+            textual_hdrs.append(f)
```
keith, line 42: this might pull in too much, depends on what can be passed to hdrs in this rule. maybe we should be doing some extension validation in the rule's attr?

cgrindel (approved):

cgrindel:
@mergifyio queue

keith:
@cgrindel can you check this one, i think it didn't merge because of an unresolved comment

cgrindel:
I cut [a new release](https://github.com/bazelbuild/bazel-central-registry/pull/8796) now that this is merged.

keith:
thanks!

PR #2294, opened 2026-05-16
fix: ensure that lowercased `.s` files are collected

adincebic:
My attempt to integrate [InjectionNext](https://github.com/johnno1962/InjectionNext) was failing with linker errors due to missing assembly files.
Prior to this change we were collecting `.S` but not also the lowercase variant `.s`.
I did consider approach where we would lowercase the extension but ended up just adding the `.s` to the list since it is explicit that people tent to use both `.s` and `.S`.


cgrindel (approved):

cgrindel:
@mergifyio queue

PR #2315, opened 2026-05-27
fix: download binaries with custom BUILD file

keith:
If you provide a custom BUILD file for a target, you likely still need
its binary targets to be downloaded like normal

PR #2316, opened 2026-05-27
fix: Separate artifact downloading from build file generation

john-flanagan:
Separate downloading of remote artifacts from build file generation. This ensures that remote artifacts are downloaded even when a custom build file is provided.

Alternate approach to https://github.com/cgrindel/rules_swift_package_manager/pull/2315 as discussed on Slack:
https://bazelbuild.slack.com/archives/C09TMF7SK5M/p1779902023383319?thread_ts=1779467061.166369&cid=C09TMF7SK5M

### DIFF
--- a/swiftpkg/internal/local_swift_package.bzl
+++ b/swiftpkg/internal/local_swift_package.bzl
@@ -81,6 +81,9 @@ def _local_swift_package_impl(repository_ctx):
         cached_json_directory,
         target_deps = repository_ctx.attr.target_deps,
     )
+
+    repo_rules.download_artifacts(repository_ctx, pkg_ctx)
+
     repo_rules.gen_build_files(repository_ctx, pkg_ctx)
 
     return update_attrs(repository_ctx.attr, _ALL_ATTRS.keys(), {})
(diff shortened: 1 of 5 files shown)


cgrindel (approved):
LGTM. I did not review the entire Slack thread. However, it looks like this PR supersedes #2315. Correct?

cgrindel:
@mergifyio queue

john-flanagan:
@cgrindel This PR or https://github.com/cgrindel/rules_swift_package_manager/pull/2315 both accomplish roughly the same thing. @Keith and I were working through it in that thread.

My two cents is that I like this approach of separating the two steps from each other rather than coupling the archive downloading to the build file generation, but either PR solves the issue.

Your call between this PR or https://github.com/cgrindel/rules_swift_package_manager/pull/2315, but we don't need both.

cgrindel:
@mergifyio queue

keith:
can we rerun CI? 502 from a url

cgrindel:
> can we rerun CI? 502 from a url

Re-running now.

PR #2319, opened 2026-05-28
fix: only attach Apple resource bundles on Apple platforms

luispadron:
## Problem

SwiftPM packages that declare resources (e.g. SwiftProtobuf's `PrivacyInfo.xcprivacy`) get an `apple_resource_bundle` / `apple_precompiled_resource_bundle` generated and attached to the target's `data` **unconditionally**.

rules_apple resource bundles require a resolved Apple platform constraint, so building such a package for a non-Apple platform (e.g. Linux) fails during analysis:

```
in environment_plist rule @@rules_apple+//apple/internal:environment_plist_ios:
Error in fail: ERROR: A valid Apple platform constraint could not be found from the resolved toolchain.
```

This blocks any Linux consumer whose dependency graph includes a resource-bearing SwiftPM package.

## Fix

Gate the resource-bundle `data` attachment behind apple_support's `apple` condition via `select()`, for both the Swift and Clang target generators:

```starlark
data = select({
    "@apple_support//configs:apple": [":Foo.rspm_resource_bundle"],
    "//conditions:default": [],
}),
```

Non-Apple targets get no bundle dependency. The `Bundle.module` accessor stays in `srcs` on all platforms, so the package's Swift/ObjC sources still compile; the Apple-only bundle is simply not pulled into the non-Apple graph.

## Testing

- Updated the golden expectations in `swiftpkg_build_files_tests.bzl` (file-path resource, precompiled bundle, and ObjC resource cases).
- `bazel test //swiftpkg/internal/... //swiftpkg/tests/...` — 251/251 pass.

### DIFF
--- a/swiftpkg/internal/swiftpkg_build_files.bzl
+++ b/swiftpkg/internal/swiftpkg_build_files.bzl
@@ -215,9 +215,9 @@ def _swift_target_build_file(repository_ctx, pkg_ctx, target):
             target,
         )
         all_build_files.append(swift_apple_res_bundle_info.build_file)
-        _update_attr_list("data", ":{}".format(
+        attrs["data"] = _apple_resource_bundle_data(
             swift_apple_res_bundle_info.bundle_label_name,
-        ))
+        )
         _update_attr_list("srcs", ":{}".format(
             swift_apple_res_bundle_info.accessor_label_name,
         ))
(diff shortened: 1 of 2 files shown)


cgrindel (approved):
Looks reasonable to me.

luispadron:
@mergify queue

jszumski (approved):

PR #2320, opened 2026-05-28
feat: use feature for Swift 6 instead of copts

keith:
rules_swift is special casing this feature and making other decisions
based on if it is enabled. Without this the compiler produces a warning
@ rules_swift HEAD since it assumes you're building for Swift 5.


cgrindel (approved):

cgrindel:
@mergifyio queue

PR #2344, opened 2026-06-09
fix: find swift via xcrun or PATH

luispadron:
Previously this used the `swift_worker` and had a fallback for linux to use `PATH`, the `swift_worker` logic breaks in rules_swift 4.0+ due to response file usage and was really just a proxy for `xcrun --find swift` anyway.


cgrindel (approved):

cgrindel:
@mergifyio queue

PR #2345, opened 2026-06-09
fix: stop deleting modulemaps in library based xcframeworks

keith:
xcframework directories can vendor direct static libraries with headers.
In this case they don't match the `*.framework` glob, but we still need
to keep the modulemaps. Example valid layout:

```
IDKitFFI.xcframework/macos-arm64_x86_64/Headers/IDKit/idkit_coreFFI.h
IDKitFFI.xcframework/macos-arm64_x86_64/Headers/IDKit/module.modulemap
IDKitFFI.xcframework/macos-arm64_x86_64/libidkitFFI.a
```

rules_apple then handles this modulemap correctly.

### DIFF
--- a/swiftpkg/internal/repo_rules.bzl
+++ b/swiftpkg/internal/repo_rules.bzl
@@ -258,6 +258,8 @@ def _remove_modulemaps(repository_ctx, directory, targets):
         directory,
         "module.modulemap",
         exclude_paths = [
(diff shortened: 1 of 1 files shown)


brentleyjones (approved):

keith:
ci flake it looks like

PR #2365, opened 2026-06-19
Support localhost HTTP registries without checksum

### LINKED ISSUE #2006: SPM registry hosted on `http` not working
I have an internal SPM registry hosted on `http` endpoint and it seems to be working fine with `swift package` but doesn't work with the rule. I am getting the following error while building my target.

```
...registry_swift_package.bzl in _download_and_parse_package_json
  repository_ctx.download(

Error in download: java.io.IOException: No URLs left after removing plain http URLs due to missing checksum. Please provide either a checksum or an https download location.
```

So, is there a way to get it work with `http` endpoint of the registry?

raiyyan089:
## Problem

When using a localhost HTTP registry (e.g. a local proxy that mirrors a remote Swift package registry), `registry_swift_package` fails at the metadata download step:

```
Error in download: java.io.IOException: No URLs left after removing plain http URLs due to missing checksum.
Please provide either a checksum or an https download location.
```

This happens because `_download_and_parse_package_json` calls `repository_ctx.download()` without a checksum — which is correct, since the checksum is contained *inside* the metadata response and cannot be known ahead of time. However, Bazel rejects plain `http://` URLs without checksums as a security measure (bazelbuild/bazel#8607).

This makes it impossible to use `registry_swift_package` with a localhost registry proxy (e.g. `http://localhost:8080/registry/...`).

## Fix

Detect `http://localhost` and `http://127.0.0.1` URLs and fall back to `repository_ctx.execute()` with `curl` for the metadata fetch. This is safe because:

- Traffic never leaves the machine
- The subsequent source archive download already has a checksum (extracted from the metadata response) and continues to use `repository_ctx.download_and_extract()` normally
- `curl` is universally available on macOS and Linux build environments

The existing behavior for all non-localhost URLs is completely unchanged.

## Use Case

Local registry proxies that cache/mirror remote Swift package registries for offline builds, network-restricted environments, or corporate build infrastructure.

Fixes #2006

### DIFF
--- a/swiftpkg/internal/registry_swift_package.bzl
+++ b/swiftpkg/internal/registry_swift_package.bzl
@@ -60,6 +60,10 @@ def _get_registry_url(*, id, registries):
     """
     return registries.scoped.get(id.scope, registries.default)
 
+def _is_localhost_url(url):
+    """Returns True if the URL points to a localhost address."""
+    return url.startswith("http://localhost") or url.startswith("http://127.0.0.1")
+
 def _download_and_parse_package_json(*, id, output, repository_ctx, registry_url, version):
     """Downloads and parses the package JSON metadata from the registry.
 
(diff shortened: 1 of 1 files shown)


cgrindel (changes requested):
Thank you for the submission! 

I am concerned about adding an implicit dependency on `curl` for this to work. There are well-founded reasons why Bazel does not support HTTP for downloads. I would encourage any download to be done using a secure protocol.

@luispadron @brentleyjones @jpsim @adincebic Any thoughts?

If we were to move forward with something like this, I think there are a few things that we would need to update:

- Introduce a new flag called `allow_insecure_http` or something similar that defaults to `False`. We would only proceed down the insecure download code path if the flag was explicitly set to `True`.
- Check for the presence of `curl` and fail gracefully if it is not present.

raiyyan089:
Closing this — the fix should be upstream in Bazel itself (exempting localhost from the http+checksum requirement in repository_ctx.download()) rather than working around it here. Thanks for the review feedback.

PR #2367, opened 2026-06-23
fix: Ignore duplicate modulemaps at the same path

keith:
Unlike all the other branches in this loop modulemaps are not allowed to
be duplicated. The callers of this function can pass a list with
duplicate entries. For modulemaps we don't need to error unless there
are different files.

Fixes https://github.com/cgrindel/rules_swift_package_manager/pull/1452


cgrindel (approved):

cgrindel:
@mergifyio queue

PR #2370, opened 2026-06-23
fix: read module name from modulemap when relevant

keith:
Previously if a user defined a target named 'foo' and in their custom
modulemap they had 'module bar', the module would be compiled with the
wrong module name and fail. The modulemap module name now takes
priority.


cgrindel (approved):

cgrindel:
@mergifyio queue

PR #2374, opened 2026-06-24
fix: Replace the bare repo working directory in stored JSON files

john-flanagan:
The stored `desc.json`, `dump.json`, and `pkg_info.json` all contain the bare repository working directory inside of a JSON string literal. Before this PR those paths were left in the stored JSON files which causes the repo content to not be reproducible between workspaces or across machines.

This change ensures that the the repo content cache is reproducible and that the cached JSON files don't contain absolute paths when using the `cached_json_directory` feature.


erikkerber:
Verified this works in our repo.

erikkerber (approved):

brentleyjones (approved):

cgrindel (approved):

PR #2383, opened 2026-06-30
fix: issue where module map parser fails to parse nested comments

### LINKED ISSUE #2382: Module map parser fails on module maps with comments inside module bodies
The module map parser fails when a `module.modulemap` file contains `//` or `/* */` comments inside a module body. For example, this valid module map:

```
framework module MyLib {
    umbrella header "MyLib.h"

    // Private API
    explicit module Internal {
        header "Foo.h"
    }
}
```

produces the error:

```
Unexpected prefix tokens found before end of line. tokens: [struct(type = "comment", value = "// Private API")]
```

The top-level parser in `parser.bzl` already skips comment tokens, but the module member collector in `collect_module_members.bzl` does not. Comment tokens fall through to the `else` branch, accumulate as unrecognized prefix tokens, and then trigger an error when the next newline is encountered.

This affects any Swift package whose `module.modulemap` contains comments between declarations — for example, [ably-cocoa](https://github.com/ably/ably-cocoa/blob/main/Source/Ably.modulemap).

daltonclaybrook:
Resolves #2382


adincebic (commented):
This looks reasonable to me. I would appreciate if you add the Swift package to examples which would serve as an integration test.

adincebic (approved):

daltonclaybrook:
> This looks reasonable to me. I would appreciate if you add the Swift package to examples which would serve as an integration test.

@adincebic Done!

adincebic:
@cgrindel

cgrindel (approved):

cgrindel:
Thank you!

cgrindel:
@mergifyio queue

PR #2397, opened 2026-07-11
feat: support module aliases for disambiguation

congt:
Swift Package Manager lets a target dependency rename a product's modules
via `moduleAliases` ([SE-0339](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0339-module-aliasing-for-disambiguation.md)). Read those declarations from the root
package manifest and apply them, scoped to the package that provides the
aliased product:

- pkginfos: retain `moduleAliases` (index 2 of the dump JSON product
  tuple) on product references.
- swift_deps extension: resolve each alias declaration to the package
  identity that provides the product, pass that package its own renames
  via `module_aliases`, and pass every generated repository (source
  control, local, and registry) the full identity-keyed map via
  `dep_module_aliases` so it can derive the flags for its direct
  dependencies.
- pkg_ctxs: compute the `-module-alias` flags for a package as its own
  renames plus the renames of the package identities it directly depends
  on (per SE-0339, a package only sees modules of packages it declares
  directly; the renaming package needs the flag too because sources
  self-qualify with their own module name).
- swiftpkg_build_files: a Swift target is renamed only when its own
  package aliases it; a same-named module in an unrelated package keeps
  its name. All Swift targets in affected packages get the
  `-module-alias` copts so package sources keep importing the original
  name.

Bazel targets outside the package graph import the replacement name.
This resolves module name collisions between a package module and a
non-package module in the same build, which is otherwise fatal under
rules_swift's `swift.use_explicit_swift_module_map` (one entry per
module name, so one module silently shadows the other).

Limitations (documented on the attrs): aliases are read from the root
manifest only, and per SE-0339 only pure-Swift source modules can be
aliased.


cgrindel:
FYI. I started looking at this PR. Overall, it looks good. I asked Claude to flesh out the work by adding some unit tests. I will post my feedback shortly.

cgrindel:
Hi @congt — nice work on this! I pushed a commit (583bcb90) adding test coverage for the cross-package module-alias propagation path, which wasn't exercised yet. Hope you don't mind the direct push; happy to adjust anything.

**What I added:**

- **Extended `module_aliases_example`** so it actually exercises propagation to a direct dependent. `swift-game` now depends on `swift-log` and its source writes `import Logging`. Since the root manifest renames that module to `SwiftLog`, `swift-game` only compiles once it receives the propagated `-module-alias Logging=SwiftLog` flag — so the example is now a build-time regression guard for the `dep_module_aliases` path (previously every built package in the example was dependency-free, so that branch never ran).
- **`swiftpkg/tests/pkg_ctxs_tests.bzl`** (new) — covers `pkg_ctxs.new`'s `module_alias_flags`: own aliases, dependency-propagated aliases, non-dependency identities ignored, and the merge of both.
- **`module_aliases_test`** in `swiftpkg_build_files_tests.bzl` — covers the emission side: the `module_name` rename and the `-module-alias` copts for both own and propagated aliases.

`Package.resolved` was left untouched — `swift-log` was already pinned (1.14.0) as a root dependency and `swift-game`'s `from: "1.5.0"` is satisfied by it, so no re-resolution was needed.

Verified locally: `bazel test //swiftpkg/tests/...` → 195/195 pass (incl. bzlformat lint), and the `module_aliases_example` integration test passes.

cgrindel:
@mergifyio queue

cgrindel (approved):

cgrindel:
@mergifyio queue

cgrindel:
@mergifyio queue

PR #2412, opened 2026-07-18
fix(swiftpkg): store portable local package paths in the lock file

### LINKED ISSUE #2405: Path dependencies write absolute path in the Bazel lock file
Using `path:` dependencies is the Packages.swift causes the Bazel lock file to contain absolute path and make the lock file non-portable.

Something like:

```swift
.package(path: "ThirdParty/foo"),
```

Renders as the following in the Bazel lock file:

```
  "swiftpkg_ foo": {
    "repoRuleId": "@@rules_swift_package_manager+//swiftpkg/internal:local_swift_package.bzl%local_swift_package",
    "attributes": {
      "bazel_package_name": "swiftpkg_foo",
      "env": {
        "ENABLE_GPU_SUPPORT": "1"
      },
      "env_inherit": [],
      "path": "/checkout_dir/ThirdParty/foo",    ‼️ this is an absolute path.
      "cached_json_directory": "",
      "target_deps": {}
    }
  },
```

This using Bazel version 8.7.0 or 9.2.0

cgrindel:
Fixes #2405.

Local `.package(path:)` dependencies were writing an **absolute** path into `MODULE.bazel.lock` (as the `local_swift_package` repo rule's `path` attribute), so committing the lock file broke it for every other machine and checkout:

```json
"swiftpkg_foo": {
  "repoRuleId": ".../local_swift_package.bzl%local_swift_package",
  "attributes": {
    "path": "/checkout_dir/ThirdParty/foo",   // ‼️ absolute, machine-specific
    ...
  }
}
```

- The `swift_deps` module extension now relativizes the local package path against the Bazel workspace root before declaring `local_swift_package`, so the stored value becomes e.g. `third_party/foo`.
- The workspace root is derived from the `swift` label rather than the `Package.swift` directory, so it stays correct when `Package.swift` lives in a subdirectory (e.g. `swift = "//:swift/Package.swift"` with `../third_party/...` deps).
- No change needed on the rule side — `local_swift_package` already re-absolutizes a relative `path` against `repository_ctx.workspace_root` at fetch time.
- Local packages that resolve *outside* the workspace root can't be made portable, so they're left absolute and emit a warning.
- Added a `relativize_repo_path` helper in `repository_utils` with unit tests (under-root, subdir prefix, sibling-prefix guard, outside-root passthrough, exact-root, trailing slash, empty root).

Verified against `language_modes_example` (paths now `third_party/swift_5_package`, full build green) and `resources_example` (subdir `Package.swift` + `../third_party/...` deps relativize to `third_party/...` against the Bazel root). Full `bazel test //...` passes (399/399).

https://claude.ai/code/session_01Fa4ZpqwGKjHyUibpwwq9xQ


daltonclaybrook:
Nice! I'm eager to get this merged as I'm running into this issue as well.

adincebic (approved):

PR #2418, opened 2026-07-21
fix: support versioned macOS frameworks

luispadron:
## Why
The importer only checks `<Name>.framework/<Name>` when determining link type, so it can fail before `rules_apple` processes a documented versioned macOS framework layout.

## What
- Fall back to `Versions/Current/<Name>` when the root framework executable is not discoverable
- Add regression coverage for dynamic versioned macOS frameworks

## References
- Apple bundle guidance: https://developer.apple.com/documentation/bundleresources/placing-content-in-a-bundle
- Framework anatomy: https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/FrameworkAnatomy.html
- `rules_apple` versioned import support: https://github.com/bazelbuild/rules_apple/pull/2133
- `bazel test //swiftpkg/...` (270 tests passed)

Generated with Codex

### DIFF
--- a/swiftpkg/internal/artifact_infos.bzl
+++ b/swiftpkg/internal/artifact_infos.bzl
@@ -75,7 +75,7 @@ def _new_framework_info_from_framework_dir(repository_ctx, path):
     """
     framework_name = _framework_name_from_path(path)
 
-    # Frameworks have a structure like the following:
+    # Unversioned frameworks have a structure like the following:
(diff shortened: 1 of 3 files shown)


cgrindel (approved):

cgrindel:
@mergifyio queue

PR #2424, opened 2026-07-25
fix(swiftpkg): determine link type from Mach-O headers instead of file(1)

cgrindel:
## Why

Renovate PRs have been failing their post-upgrade task:

```
Error in fail: Failed to determine the file type for .../FirebaseAnalytics.framework/FirebaseAnalytics
stderr: "execvp(file, ...)": No such file or directory
```

The xcframework importer shelled out to `file(1)` to decide whether a framework binary is static or dynamic. The self-hosted Renovate container has no `file`, and the `otool` fallback is Xcode-only, so it could never have worked on Linux either.

## What

- Read the Mach-O magic bytes and the `filetype` field directly instead of parsing `file` output. Bytes come from `od` (coreutils, already assumed by the neighboring `find`/`cp` calls), and only the first 24 bytes of the header are read: one `od` spawn for a thin binary, two for a universal one.
- `MH_DYLIB` and `MH_DYLIB_STUB` are the file types that carry `LC_ID_DYLIB`; everything else is static.
- Removed `repository_files.file_type` and the `otool` fallback.
- Moved `link_types` to its own module so `mach_o.bzl` can use it without a dependency cycle; `artifact_infos` re-exports it, so no callers changed.

## Behavior changes

- Non-Mach-O input now fails with "Unrecognized magic bytes" instead of silently resolving to `static`.
- Truncated or malformed fat headers are reported accurately.
- For a universal binary with mixed slice kinds, the first slice decides; the old code let any archive slice win.

## Tests

Three layers: `mach_o_tests` covers the header formats synthetically (thin/fat, 32/64-bit, both endiannesses, `fat64`); `repository_files_tests` covers `od` output parsing (GNU and BSD spacing, wrapped rows); and `verify_fixtures_test` re-reads seven checked-in real Mach-O fixtures with the real `od`, on both `ubuntu-22.04` and `macos-15` in CI. Each fixture's expected link type is derived from `file`/`otool` rather than from a second walk of the header, so it is a real equivalence test. Fixture details are in `swiftpkg/tests/fixtures/mach_o/README.md`.

Verified end to end with `//:update_build_files` in `examples/shake_ios_example`, which fetches the real `Shake.xcframework` and still resolves to `apple_dynamic_xcframework_import`.


cgrindel:
@mergifyio queue

PR #2430, opened 2026-07-28
fix: anchor swift package tool state paths to the workspace directory

luispadron:
### What

Anchors the `swift_package_tool` targets' `build_path`/`cache_path`/`config_path`/`security_path` to `BUILD_WORKSPACE_DIRECTORY` when they are relative, matching the existing `package_path` behavior. Absolute paths pass through unchanged.

### Why

Relative paths resolve against the process working directory, which under `bazel run` is the target's runfiles tree in the output base. Keeping SPM state there has two problems:

1. **It races with Bazel itself.** Bazel releases its command lock before exec'ing a `bazel run` binary, so an unrelated bazel command can rebuild or delete the runfiles tree while `swift package resolve` is still running against it. The state is also configuration-specific (one `.build` per output config) and wiped by `bazel clean`.
2. **SwiftPM's inter-process locking never engages.** SwiftPM serializes concurrent invocations via a file lock in the scratch directory. With the scratch dir buried in the runfiles tree, a direct `swift package resolve --package-path <pkg>` (default scratch: `<pkg>/.build`) and a `bazel run @swift_package//:resolve` use different scratch dirs, hold different locks, and can concurrently rewrite the same `Package.resolved`. Anchored to the workspace (e.g. configured to `<pkg>/.build`), both flavors contend on the same lock and serialize.

We hit both in a large monorepo as intermittent resolve/lint flakes whenever git hooks, CI, and developers raced package resolution.


cgrindel:
@luispadron Is the infinite symlink expansion error related to this change?

cgrindel:
@luispadron I will hold off reviewing for now. Ping me when it is ready to review.

adincebic (approved):

cgrindel (approved):
I hate that we have to write back to the Bazel workspace. However, I think the reasoning makes sense. I guess this is the price that we pay to keep Bazel and Swift happy. 

Thanks for putting this together!

cgrindel:
@mergifyio queue

cgrindel:
@mergifyio dequeue

swiftpkg/internal/swift_package_lib.sh
```diff
@@ -177,6 +196,14 @@ spl_run_swift_package() {
   # Resolve package_path relative to workspace.
   package_path="${BUILD_WORKSPACE_DIRECTORY}/${package_path}"
```
cgrindel: The PR description says this matches the existing `package_path` behavior, but it doesn't quite: this line anchors unconditionally and isn't absolute-aware. Passing an already-absolute value yields `<workspace>//already/abs`. In practice `package_path` is always label-derived and relative, so it isn't a live bug — but routing it through `spl_anchor_to_workspace` too would make the claim true and remove the divergence between the two code paths.

swiftpkg/tests/swift_package_lib_tests/swift_package_lib_test.sh
```diff
@@ -199,6 +199,21 @@ assert_argv_has "--replace-scm-with-registry" \
...
+assert_argv_has "${workspace_dir}/.build" \
+  "relative build_path should anchor to BUILD_WORKSPACE_DIRECTORY"
+assert_argv_has "${workspace_dir}/.cache" \
+  "relative cache_path should anchor to BUILD_WORKSPACE_DIRECTORY"
+assert_argv_has "${workspace_dir}/.security" \
+  "relative security_path should anchor to BUILD_WORKSPACE_DIRECTORY"
+assert_argv_lacks ".build" \
+  "unanchored relative build_path should not appear"
+
+# Absolute state paths must pass through unchanged.
```
cgrindel: These assertions are valid — I confirmed `assert_argv_lacks` uses `grep -qxF`, so `assert_argv_lacks ".build"` doesn't spuriously match `${workspace_dir}/.build`.

The gap is the one case with a real on-disk side effect: a **relative** `config_path` combined with `--registries_json`. `spl_setup_registries` now does `mkdir -p "<workspace>/.config"` and drops a `registries.json` symlink pointing into the Bazel output base — new persistent state in the user's source tree, and a symlink that dangles after `bazel clean` (it self-heals on the next run since `ln -sf` re-runs). The existing registries test at lines 320-356 passes an **absolute** `config_path`, so nothing exercises the anchored path.

Smaller additions worth having: a direct unit test of `spl_anchor_to_workspace` (absolute, relative, `./`-prefixed, value-with-space); an absolute `build_path` case, since only `config_path` covers pass-through today; and a `workspace_dir` containing a space (`workspace_dir="${run_dir}/ws"` has none — this does work, it's just untested).

docs/faq.md
```diff
@@ -35,6 +36,24 @@ commands. These commands result in a `.build` directory being created. The direc
...
 
+## Where do the `swift_package_tool` targets (e.g. `@swift_package//:resolve`) write their state?
+
+Into your workspace. The `build_path`, `cache_path`, `config_path` and `security_path` attributes
+default to `.build`, `.cache`, `.config` and `.security`; a relative value is resolved against
+`BUILD_WORKSPACE_DIRECTORY`, matching `package_path`. This keeps Swift package manager state out of
+the runfiles tree, where Bazel may rebuild or delete it mid-command, and lets a direct
+`swift package resolve` share Swift package manager's file locks with the Bazel targets.
+
+Because `.build` holds package checkouts that carry their own `BUILD` files, add these directories
```
cgrindel: This is the right advice, but `.gitignore` in this repo only covers `**/.build` — `.cache`, `.config` and `.security` aren't there, nor are the `spm-*` variants the `interesting_deps` example uses. Running any of these examples locally now leaves untracked directories behind. Worth updating the repo's own `.gitignore` in this PR alongside telling users to update theirs.

examples/interesting_deps/.bazelignore
```diff
@@ -0,0 +1,4 @@
+.build
```
cgrindel, line 1: These four entries don't match what this example produces. `examples/interesting_deps/MODULE.bazel:78-84` configures non-default state paths:

```
build_path = "spm-build",
cache_path = "spm-cache",
config_path = "spm-config",
security_path = "spm-security",
```

So after this change the example writes `spm-build/`, `spm-cache/`, `spm-config/` and `spm-security/` into its workspace root, and every entry here is dead. `do_test:16` runs `bazel run @swift_package//:resolve` before `bazel test //...`, which makes this the one example where the ignore file is actually load-bearing.

```suggestion
spm-build
spm-cache
spm-config
spm-security
```

swiftpkg/internal/swift_package_lib.sh
```diff
@@ -41,6 +41,25 @@ spl_setup_registries() {
...
+# `swift package` invocations against the same package share SwiftPM's
+# file locks with these targets.
+#
+# Arguments:
+#   $1 - the path to anchor (may be absolute)
+spl_anchor_to_workspace() {
+  local path="$1"
+  case "${path}" in
+    /*) printf '%s' "${path}" ;;
+    *) printf '%s' "${BUILD_WORKSPACE_DIRECTORY}/${path}" ;;
```
cgrindel: `${BUILD_WORKSPACE_DIRECTORY}` has no `:-` or `:?` default here. Called through `spl_run_swift_package` this is unreachable — lines 171-175 guard it first — but this is a sourced library and the sibling `spl_setup_registries` is called directly by the tests, so a standalone call under `set -u` dies with a bare `unbound variable`. `${BUILD_WORKSPACE_DIRECTORY:?BUILD_WORKSPACE_DIRECTORY must be set}` would make that self-describing.

Two input shapes also pass through unchanged, and both got worse with this change rather than better:

- `~/.spmbuild` yields `<workspace>/~/.spmbuild` — a literal `~` directory. Previously that landed in the throwaway runfiles tree; now it lands in the user's source tree.
- `../shared-cache` yields `<workspace>/../shared-cache`, silently writing outside the workspace.

Both were pre-existing, so neither has to be fixed here. But rejecting a leading `~` with a clear error is cheap, and `..` is worth either documenting as supported or rejecting.

swiftpkg/internal/swift_package_lib.sh
cgrindel: The PR description says this matches the existing `package_path` behavior, but it doesn't quite: this line anchors unconditionally and isn't absolute-aware. Passing an already-absolute value yields `<workspace>//already/abs`. In practice `package_path` is always label-derived and relative, so it isn't a live bug — but routing it through `spl_anchor_to_workspace` too would make the claim true and remove the divergence between the two code paths.

swiftpkg/tests/swift_package_lib_tests/swift_package_lib_test.sh
cgrindel: These assertions are valid — I confirmed `assert_argv_lacks` uses `grep -qxF`, so `assert_argv_lacks ".build"` doesn't spuriously match `${workspace_dir}/.build`.

The gap is the one case with a real on-disk side effect: a **relative** `config_path` combined with `--registries_json`. `spl_setup_registries` now does `mkdir -p "<workspace>/.config"` and drops a `registries.json` symlink pointing into the Bazel output base — new persistent state in the user's source tree, and a symlink that dangles after `bazel clean` (it self-heals on the next run since `ln -sf` re-runs). The existing registries test at lines 320-356 passes an **absolute** `config_path`, so nothing exercises the anchored path.

Smaller additions worth having: a direct unit test of `spl_anchor_to_workspace` (absolute, relative, `./`-prefixed, value-with-space); an absolute `build_path` case, since only `config_path` covers pass-through today; and a `workspace_dir` containing a space (`workspace_dir="${run_dir}/ws"` has none — this does work, it's just untested).

swiftpkg/tests/swift_package_lib_tests/swift_package_lib_test.sh
```diff
@@ -199,6 +199,21 @@ assert_argv_has "--replace-scm-with-registry" \
...
+assert_argv_has "${run_dir}/.config" \
+  "absolute config_path should pass through unchanged"
+
```
cgrindel, line 216: Claude noticed a gap in the tests. It suggested adding the following. 

```suggestion
# MARK - spl_run_swift_package relative config_path + --registries_json

# With a relative config_path, the registries symlink must land under
# BUILD_WORKSPACE_DIRECTORY. Guards against the anchoring being dropped
# or reordered after spl_setup_registries.
rel_cfg_dir="$(new_tmp_dir)"
rel_cfg_file="${rel_cfg_dir}/registries.json"
echo '{"registries":{}}' >"${rel_cfg_file}"
: >"${swift_args_file}"

PATH="${run_dir}:${PATH}" BUILD_WORKSPACE_DIRECTORY="${workspace_dir}" \
  spl_run_swift_package \
  --cmd resolve \
  --package_path pkgsub \
  --build_path .build \
  --cache_path .cache \
  --config_path .config \
  --security_path .security \
  --enable_build_manifest_caching true \
  --enable_dependency_cache false \
  --manifest_cache shared \
  --replace_scm_with_registry false \
  --use_registry_identity_for_scm false \
  --registries_json "${rel_cfg_file}"

assert_argv_has "${workspace_dir}/.config" \
  "relative config_path should anchor to BUILD_WORKSPACE_DIRECTORY"

rel_cfg_symlink="${workspace_dir}/.config/registries.json"
if [[ ! -L ${rel_cfg_symlink} ]]; then
  fail "expected registries symlink at ${rel_cfg_symlink}"
fi
assert_equal \
  "$(readlink -f "${rel_cfg_file}")" "$(readlink "${rel_cfg_symlink}")" \
  "relative config_path should anchor the registries symlink to the workspace"
```

cgrindel (changes requested):
Claude noticed a couple things after my approval. We should be good to go after these are addressed.

PR #2369, opened 2026-06-23
fix: don't include everything under public includes

keith:
Previously all source files under the public headers path would be
included. This is different from SwiftPM which only includes source
files from public headers if they are under the sources directory.

PR #2432, opened 2026-07-29
fix(swiftpkg): discover sibling public headers for file-level sources

ma-oli:
When a C/Objective-C target names an individual source file in its manifest `sources` (e.g. `sources: ["util.c"]`) and its public headers path is the directory containing that file, sibling public headers were no longer discovered after #2369.

`clang_files.public_include_file_scan_paths` returned the source file itself as the scan root. Listing a file yields only that file, so sibling headers (e.g. `util.h`) in the same public-include directory were dropped from `hdrs`.

Resolve any file-valued source path under the public include to its containing directory before scanning, so those headers are found again. The scan stays bounded to the target's own source subtrees and results are still filtered to public include files, so no unrelated sources are compiled.


swiftpkg/internal/clang_files.bzl
```diff
@@ -196,22 +197,44 @@ def _is_under_path(path, parent):
...
     include root would pull files from sibling directories outside the explicit
     source paths. If no source path sits under the public include root, treat
     the public include as a separate header directory and scan it directly.
+
+    An explicit source path may name an individual file (e.g.
+    `sources: ["util.c"]`) rather than a directory. Scanning a file yields
+    only that file, so any sibling public headers in the same directory would be
+    missed. In that case, scan the file's containing directory instead. The scan
+    remains bounded to the target's own source subtrees, and the caller filters
+    the results to public include files, so no unrelated sources are compiled.
```
cgrindel, line 213: Is this codified in Swift Package Manager as documented behavior?

cgrindel (commented):
Overall, LGTM. However, I just want to confirm that this matches SPM behavior.

swiftpkg/internal/clang_files.bzl
cgrindel: Yes, on both counts.

Documented in [CreatingCLanguageTargets.md](https://github.com/swiftlang/swift-package-manager/blob/main/Sources/PackageManagerDocs/Documentation.docc/CreatingCLanguageTargets.md): "Swift package manager uses the `include` directory as an umbrella directory; all headers under it are included in the module." No mention of `sources` — the public header surface is defined by the directory layout.

Codified in `TargetSourcesBuilder.computeRule()`, which filters *only* the `.compile` rule when `sources` is declared:

```swift
// Don't automatically match compile rules if target's sources are
// explicitly declared in the package manifest.
if declaredSources != nil {
    return rules.filter { $0.rule != .compile }
}
```

Headers still match automatically anywhere under the target path, `computeContents()` walks the whole target directory rather than the declared sources, and `ModuleMapGenerator` reads the public headers dir directly without consulting `sources`. So declaring `sources:` bounds compilation only.

If anything this PR is still narrower than SwiftPM — it only re-expands to the named file's containing directory. Happy to keep it conservative here, but worth noting the `is_public_include_file` filter at `pkginfos.bzl:1520` is what actually prevents compiling unrelated sources, independent of the scan paths.

cgrindel (approved):
I had Claude do some research. Thanks for the contribution!

cgrindel:
@mergifyio queue

cgrindel:
@mergifyio queue

cgrindel:
@mergifyio queue

Issue #1839, opened 2025-09-25
Support for suppressing warnings in specific packages


watt:
There's a strategy [discussed on the bazelbuild community slack](https://bazelbuild.slack.com/archives/C05NZSHE7FU/p1743110146228039?thread_ts=1741113928.824389&cid=C05NZSHE7FU) for suppressing warnings in external dependencies that works like this:

> Add these lines to .bazelrc to disable warnings both remote and local
> ```
> build --features=suppress_warnings
> build --features=swift.suppress_warnings
> ```
> then add these lines to REPO.bazel to re-enable warnings for in-repo files
> ```
> repo(
>     features = [
>         "-suppress_warnings",
>         "-swift.suppress_warnings",
>     ],
> )
> ```

This works, but is an all-or-nothing approach.

One can also disable warnings on a specific repo by using a patch, but this is cumbersome to create. It would be nice to be able to suppress warnings at a package level more conveniently, perhaps using `configure_package`.

cgrindel:
> I'm leaning towards adding swift_features both globally and per-package -- adding a flag just for suppress_warnings feels like it could lead to an unnecessarily complicated API if more features are ever requested, and additional_swift_copts feels a bit low-level when the concept of features already exists.

I agree. Thanks for taking up the mantle.

jschear:
Great, I'll get to work on adding `swift_features` to `from_package` (for global configuration) and `configure_package` (for  per-package configuration).

I also just realized https://github.com/bazelbuild/rules_swift/blob/master/doc/rules.md#swift_package_configuration exists, does anyone know if that can be used to configure packages outside the root module?

PR #2439, opened 2026-08-04
feat: add target-specific Swift compiler options

karim-alweheshy:
## What

Add a repeatable `swift_deps.configure_target` module-extension tag for appending Swift compiler options to one generated Swift package target, including Swift test targets.

The tag accepts:

- a package selector (package name, package identity, or generated repository name);
- a Swift target selector (raw target, public `.rspm`, or `.rspm.__impl` name);
- a required list of `swift_copts`; and
- an optional typed condition label, emitted through `select()`.

Repeated declarations for the same target and condition preserve declaration order. Configured options are rendered after manifest/generated options so normal last-option-wins compiler flags behave as requested. Source-control, local, and registry packages carry the configuration through the same repository APIs.

The implementation fails closed when a package or target selector is ambiguous, when the package/target is unknown or non-Swift, or when a complete `configure_package(build_file = ...)` override would bypass generated targets.

## Why

Today consumers can replace an entire generated package BUILD file or apply compiler flags globally, but cannot express a small target-local Swift compile condition. A real example is compiling production-only bookkeeping out of one runtime target without changing siblings or maintaining a generated BUILD snapshot.

This is related to #1839, but does not close it: that issue also discusses package-level warning suppression and feature configuration. The target-name/configuration plumbing follows the direction established by #2275.

## Correctness details

- Condition labels are canonicalized in the declaring module, so root-repository `config_setting` labels remain valid in generated external repositories.
- Selector matching is recorded before unresolved dependencies are replaced with placeholders, preserving the normal `@swift_package//:resolve` flow for a stale `Package.resolved`.
- Package aliases matching multiple dependencies and raw/generated target-name collisions are rejected instead of applying flags broadly.
- Manifest copts and configured copts are rendered as separate expressions, with configured values last even when either side contains `select()` expressions.
- Swift test targets use the same target-local configuration path as generated Swift libraries, executables, and compiler plugins. Sibling targets remain unchanged.

## Tests

- Full-suite `bazelisk test //...` on the rebased feature series with the selected Xcode SDK supplied to Bazel: **430/430 passed**.
- On the final current-main rebase, the complete `//swiftpkg/tests/...` suite, `//bzlmod:stale_resolved_test`, and all four generated-document difftests passed: **231/231**.
- Both bzlmod integration fixtures pass uncached with a clean-runner `HOME` that injects `common --config=ci`, reproducing the GitHub Actions configuration that exposed the stale-resolved fixture gap.
- `bazelisk run //:tidy_check` passes.
- A four-arm aquery fixture verifies both flags on every selected-target Swift frontend action, and no flags on disabled or sibling actions.
- The same aquery fixture passes with `--incompatible_config_setting_private_default_visibility`, proving the external-repository `select()` can see its explicitly public condition.
- Swift test-target coverage verifies configuration placement on `swift_test` and isolation from a sibling Swift library.
- The complete `bzlmod/workspace/do_test` path runs tidy, the workspace test suite, the selected/sibling enabled/disabled aquery arms, and the executable smoke test.
- Buildifier, ShellCheck, shell/Python syntax, Swift package manifests, generated documentation, and `git diff --check` pass.


cgrindel:
@karim-alweheshy I presume that this is not ready to review because it is still in draft. Let me know when it is ready.

karim-alweheshy:
@cgrindel This is ready for review now.

I rebased the feature series onto current `main`, addressed the Swift test-target gap, refreshed the generated configuration docs, and fixed the CI-only stale-resolved fixture failure by giving that nested workspace the same `.bazelrc` imports as the other integration workspace.

Verification on the rebased series includes a full **430/430** pass. On the final current-main rebase, `//swiftpkg/tests/...`, `//bzlmod:stale_resolved_test`, and all four generated-document difftests pass **231/231**. The inherited clean-runner `common --config=ci` reproduction also passes for both bzlmod fixtures.

The PR body now documents the public API, validation behavior, test-target support, and the complete verification matrix.

cgrindel:
@karim-alweheshy Did you happen to see [the design proposal](https://github.com/cgrindel/rules_swift_package_manager/discussions/2466) and [implementation PR](https://github.com/cgrindel/rules_swift_package_manager/pull/2467) that I put up to unify a couple of related requests?

PR #2446, opened 2026-08-07
feat: add `configure_package_target` tag for per-target attribute overrides

### LINKED ISSUE #2445: `alwayslink = True` on static xcframework imports breaks linking for SDKs with unreferenced objects that have undefined symbols
## Summary

`_xcframework_import_build_file` unconditionally sets `alwayslink = True` on every `apple_static_xcframework_import`, with no way to opt out per package or per target.

When an archive contains an object that nothing references and whose symbols nothing can resolve, force-loading pulls it in and the link fails. The same SDK links fine under SPM/Xcode, which does not force-load.

## Reproduction

Any Bazel target that links `FBAudienceNetwork` (e.g. transitively via `@swiftpkg_googleads_mobile_ios_mediation_meta//:MetaAdapterTarget`) fails at link time:

```
Undefined symbols for architecture arm64:
  "_XPluginsGetDataFuncOrAbort", referenced from:
      _LogDefaultLevelSocket_GetPlugin in FBAudienceNetwork[arm64][136](PluginSockets.mm.pic.o)
      ...
  "_XPluginsGetFunctionPtrFromID", referenced from:
      _LogDirectorySocket_InvokeDirectoryFunction in FBAudienceNetwork[arm64][136](PluginSockets.mm.pic.o)
  "Plugins::SocketThreadLocalScope::SocketThreadLocalScope(int)", referenced from:
      _LogDirectorySocket_InvokeDirectoryFunction in FBAudienceNetwork[arm64][136](PluginSockets.mm.pic.o)
  "Plugins::SocketThreadLocalScope::~SocketThreadLocalScope()", referenced from:
      _OUTLINED_FUNCTION_0 in FBAudienceNetwork[arm64][136](PluginSockets.mm.pic.o)
ld: symbol(s) not found for architecture arm64
```

## Root cause

`PluginSockets.mm.pic.o`, one of 430 objects in `FBAudienceNetwork.a`, leaves those four symbols undefined. Every other undefined symbol in it resolves against libc++, libSystem, or libobjc; these four have no provider in the archive or anywhere on the link line.

The object has no ObjC classes or categories and no `__mod_init_func`, and nothing else in the archive references it, so a normal link never loads it — which is why SPM/Xcode succeed. `alwayslink = True` force-loads it anyway.

This is not Bazel-specific: the same symbols break Xcode builds whenever something forces the object to load. See [facebook-ios-sdk#2488](https://github.com/facebook/facebook-ios-sdk/issues/2488) (open, labeled "out-of-scope") and [AppLovin-MAX-SDK-iOS#385](https://github.com/AppLovin/AppLovin-MAX-SDK-iOS/issues/385).

The default itself is reasonable and should stay — it preserves symbols other SDKs need. The problem is that it cannot be overridden, which leaves affected consumers with no good workaround.

## Where

`swiftpkg/internal/swiftpkg_build_files.bzl`, in `_xcframework_import_build_file`:

```python
if artifact_info.link_type == link_types.static:
    ...
    # Firebase example requires that GoogleAppMeasurement symbols are
    # passed along.
    attrs["alwayslink"] = True
```

Introduced in #1217, separate from the `swift_library` / `clang_target` defaults in #1385 that are justified by matching SPM semantics.

## Relation to existing issues

#1222 asked for a per-package mechanism to set `alwayslink` and was closed in favor of #1385 setting it unconditionally, "without requiring a per-package configuration mechanism". That reasoning does not carry over here, in two ways.

First, the direction is reversed. #1222 wanted configuration to opt *into* force-loading where rspm diverged from SPM. Here rspm is the one diverging: SPM and Xcode do not force-load this archive, so an escape hatch restores SPM parity rather than departing from it.

Second, the "infer it from what SPM tells us" approach cannot resolve this case. Nothing in a `Package.swift` manifest indicates that a prebuilt archive contains an object with unresolvable symbols; that is only visible in the binary. #1222 also concerned `swift_library` (source targets, now the rules_swift 3.0 default) rather than `apple_static_xcframework_import`, where "mimic SPM" argues against force-loading.

#2216 asks whether the explicit `alwayslink` can be dropped now that rules_swift 3.0 always-links by default. It refers to the `swift_library` site; this issue is about the separate xcframework site, where the attribute is not redundant but actively breaking.

daltonclaybrook:
This PR resolves #2445 by adding a new tag called `configure_package_target`. This tag is similar to the `configure_package` tag, but allows users to configure individual targets of packages rather than the package itself. Currently, the only configurable property is `alwayslink`. Here's an example usage:

```
swift_deps.configure_package_target(
    package_name = "swiftpkg_fbaudiencenetwork",
    target_name = "FBAudienceNetwork",
    alwayslink = False,
)
```


cgrindel:
I believe that we have addressed the underlying issue with https://github.com/cgrindel/rules_swift_package_manager/pull/2467. If you feel that there is more to address, please let me know.

PR #2467, opened 2026-08-17
feat(swiftpkg): add bazel_target_* tags to override attrs on generated targets

cgrindel:
Implements the design proposed in #2466, unifying the goals of #2439 and #2446 behind a single buildozer-inspired API for customizing RSPM-generated Bazel targets.

- Added repeatable `swift_deps` tags that each name a declaration in a generated repo (e.g. `@swiftpkg_foo//:Bar.rspm.__impl`) and the attribute to modify.
- Typed setters: `bazel_target_set_bool`, `bazel_target_set_int`, `bazel_target_set_string`, `bazel_target_set_string_list`, and `bazel_target_set_string_dict`. Each takes a natively typed `value`.
- Typed `select()` setters: `bazel_target_set_select_bool_dict`, `bazel_target_set_select_int_dict`, `bazel_target_set_select_string_dict`, and `bazel_target_set_select_string_list_dict`. Bool and int branches are strings converted at extension evaluation, since Bazel has no bool or int dict tag attribute.
- Appenders: `bazel_target_add` and `bazel_target_add_select`, list-only.
- The `target` value is a generated repository name plus a declaration name, not a Bazel label. `use_repo` aliases and the canonical `@@` prefix are rejected.
- Root module only. Fails fast on unknown repo names, unknown target names (listing the declarations that exist), duplicate setters, and combining with `configure_package(build_file = ...)`.
- No attribute allowlist. This is a use-at-your-own-risk escape hatch. A bad attribute name fails with Bazel's own error when the generated repo loads.
- Added unit tests for the new `bazel_target_mods` module and the extension-side collection/validation.
- Updated the `pkg_manifest_minimal` example to modify a generated target, with `do_test` assertions that the values land in the generated build file and that a `select()`-valued `copts` reaches every Swift compile action of that target only under the matching define, without leaking to a sibling target.
- Added a how-to doc at `docs/modify_generated_bazel_targets.md` plus a README section.

https://claude.ai/code/session_011Ga6AG5yf6M1T45YXGyg5j


swiftpkg/internal/bazel_target_mods.bzl
```diff
@@ -0,0 +1,570 @@
...
+        A `list` of modification `dict` values in application order.
+    """
+    ordered = []
+    for verbs in _VERB_GROUPS:
+        ordered.extend([mod for mod in mods if mod["verb"] in verbs])
+    return ordered
+
+def _apply_mod(attrs, mod):
+    verb = mod["verb"]
+    attr_name = mod["attr"]
```
ssarad, line 522: @cgrindel Doesn’t this lose the attribute’s type information? `_parse_value` can only produce a scalar bool, int, or string. For example, 
```
bazel_target_set(
   attr = "copts", 
   value = "-enable-library-evolution"
)
```
generates `copts = "-enable-library-evolution"` instead of the  required string list. Conversely, `set_select` always produces list-valued branches, so it cannot configure scalar attributes. The  design in #2466 proposed a typed attribute allowlist to distinguish these cases. Should that be restored here?

cgrindel: Good point. I moved away from the allowlist because we would need to keep type information for every attribute that could be modified. 🤔

Your comment got me thinking about providing type-specific attributes: `int_value`, `str_value`, `bool_value`, `list_value`, `dict_value`. However, detecting values that match the defaults for those types has some issues. I am going to think on this a bit.

cgrindel: Good catch, and thanks. Parsing the string guessed the type from the value, and it could not express a list for `set` or a scalar for `set_select`. Rather than restore the allowlist, which would require every attribute anyone wants to touch to be in a table first, the tags now declare the type. Claude implemented the change in 0bc395e2: `bazel_target_set` is replaced by `bazel_target_set_bool`, `_int`, `_string`, `_string_list` and `_string_dict`, each with a natively typed `value`, and `bazel_target_set_select` by `bazel_target_set_select_bool_dict`, `_int_dict`, `_string_dict` and `_string_list_dict`. The only conversion left is for the bool/int select branches, since Bazel has no bool or int dict tag attribute; those must be exactly `True`/`False` or an integer and fail at extension evaluation otherwise. A value of the wrong type for the target attribute remains user error, caught by Bazel when it loads the generated BUILD file. Details are in `docs/modify_generated_bazel_targets.md`.

ssarad (approved):
Thanks for addressing the comment! LGTM

karim-alweheshy (commented):
I reviewed the current head (`306a86dc`) against Bazel's label and repository-mapping behavior. Overall, the post-merge declaration patch IR, validation, deterministic ordering, and `select()` composition look sound.

I found one API-contract mismatch that I think is worth resolving before this becomes a public escape hatch. [Discussion #2466](https://github.com/cgrindel/rules_swift_package_manager/discussions/2466) describes `target` as an exact Bazel label, and the tag documentation calls it a label string, but the implementation treats its repository portion as an RSPM-generated repository key.

I reproduced this by importing the generated repository under an alias:

```starlark
use_repo(
    swift_deps,
    log_alias = "swiftpkg_swift_log",
)
```

The real root-visible Bazel label, `@log_alias//:Logging.rspm.__impl`, queries successfully, but using it in `bazel_target_add(target = ...)` fails with `unknown repositories: log_alias`. Conversely, `@swiftpkg_swift_log//:Logging.rspm.__impl` works as the modification selector even though that spelling is no longer visible as a label from the root repository.

The accepted `@@swiftpkg_swift_log//:...` form is also not Bazel's canonical repository name. The actual canonical repository at this head is `@@rules_swift_package_manager++swift_deps+swiftpkg_swift_log`, and passing that form to the modification API fails because routing expects `swiftpkg_swift_log`.

I think the smallest and clearest fix is to describe `target` as an RSPM generated-declaration selector, explicitly require `@generated_repo_name//:declaration_name`, state that `use_repo` aliases are not accepted, and remove the misleading `@@` canonical-label support. Mapping-aware target labels could remain a separate follow-up if wanted.

Everything else I exercised held up well. The focused target-modification suites passed, the full `//swiftpkg/tests:all` suite passed, and a direct Swift action matrix confirmed that a conditional `copts` value reaches every selected `Logging` frontend action without leaking to the sibling `InMemoryLogging` target. I have that small action-level regression available if it would be useful, but I do not consider it a merge blocker.

One separate housekeeping item: the PR body still advertises the old untyped `bazel_target_set` and `bazel_target_set_select` names, while the implementation and checked-in guide now expose the typed tag family.

docs/modify_generated_bazel_targets.md
daltonclaybrook, line 1: Great doc!

daltonclaybrook (approved):
LGTM

cgrindel:
@karim-alweheshy Thanks for the careful review. You are right that `target` was described as a label but treated as an RSPM repository key. Claude made the change you suggested in 7f471c0c: the tag doc and the how-to now say the repository portion is the generated repository name, that `use_repo` aliases are not accepted, and that the name does not need to be imported for a modification to apply. The `@@` prefix is now rejected with a message explaining why, rather than silently stripped, since the value is never a canonical label. Mapping-aware target labels would need the extension to see the root module's `use_repo` mapping, which extensions cannot do, so that stays out of scope. The PR body is updated to name the typed tags. The action-level regression test sounds useful; a follow-up PR would be welcome.

cgrindel:
@karim-alweheshy Follow-up on the action-level test: Claude added it to this PR in 2687e847 rather than waiting on a separate one. The `pkg_manifest_minimal` `do_test` now runs `bazel aquery` and asserts the conditional `copts` reaches every `SwiftCompile` action for `Logging` under the define, is absent without it, and never reaches `InMemoryLogging`.

adincebic (approved):

cgrindel:
@mergifyio queue

PR #2490, opened 2026-08-30
fix: add binary targets to generated public swift_library_group

adincebic:
rules_swift has a feature to detect transitive module imports. This is referred to as layering check.
We make sure that binary targets are added to publically generated `swift_library_group` so that it satisfies the layering check.

### DIFF
--- a/swiftpkg/internal/swiftpkg_build_files.bzl
+++ b/swiftpkg/internal/swiftpkg_build_files.bzl
@@ -1319,12 +1319,36 @@ def _library_product_build_file(pkg_ctx, product):
             pkginfo_target_deps.labels_for_target(pkg_ctx.repo_name, target),
         )
 
+    binary_target_deps = []
+    for target in pkg_ctx.pkg_info.targets:
+        if product.name not in target.product_memberships:
+            continue
+        for target_dep in target.dependencies:
+            dep_target_name = None
+            if target_dep.target:
+                dep_target_name = target_dep.target.target_name
+            elif target_dep.by_name:
+                dep_target_name = target_dep.by_name.name
+            if not dep_target_name:
+                continue
+            dep_target = pkginfo_targets.get(
+                targets = pkg_ctx.pkg_info.targets,
+                name = dep_target_name,
+                fail_if_not_found = False,
+            )
+            if not dep_target or dep_target.type != target_types.binary:
+                continue
+            binary_target_deps.extend(
+                pkginfo_target_deps.bzl_select_list(pkg_ctx, target_dep),
+            )
+
     if len(target_labels) == 0:
         fail("No targets specified for a library product. name:", product.name)
     deps = [
         bazel_labels.normalize(label)
         for label in target_labels
     ]
+    deps.extend(binary_target_deps)
     return build_files.new(
         load_stmts = [swift_library_group_load_stmt],
         decls = [
@@ -1332,7 +1356,7 @@ def _library_product_build_file(pkg_ctx, product):
                 swift_kinds.library_group,
                 product.name,
                 attrs = {
-                    "deps": deps,
+                    "deps": bzl_selects.to_starlark(deps),
                     "visibility": ["//visibility:public"],
                 },
             ),
--- a/swiftpkg/internal/swiftpkg_build_files.bzl
+++ b/swiftpkg/internal/swiftpkg_build_files.bzl
@@ -1,6 +1,7 @@
 """Module for creating Bazel declarations to build a Swift package."""
 
 load("@bazel_skylib//lib:paths.bzl", "paths")
+load("@bazel_skylib//lib:sets.bzl", "sets")
 load("@cgrindel_bazel_starlib//bzllib:defs.bzl", "bazel_labels", "lists")
 load(
     "//config_settings/bazel/compilation_mode:compilation_modes.bzl",
@@ -1319,7 +1320,15 @@ def _library_product_build_file(pkg_ctx, product):
             pkginfo_target_deps.labels_for_target(pkg_ctx.repo_name, target),
         )
 
-    binary_target_deps = []
+    if len(target_labels) == 0:
+        fail("No targets specified for a library product. name:", product.name)
+
+    deps = [
+        bazel_labels.normalize(label)
+        for label in target_labels
+    ]
+    target_label_set = sets.make(deps)
+    binary_target_dep_pairs = sets.make()
     for target in pkg_ctx.pkg_info.targets:
         if product.name not in target.product_memberships:
             continue
@@ -1338,15 +1347,34 @@ def _library_product_build_file(pkg_ctx, product):
             )
             if not dep_target or dep_target.type != target_types.binary:
                 continue
-            binary_target_deps.extend(
-                pkginfo_target_deps.bzl_select_list(pkg_ctx, target_dep),
-            )
+            for binary_target_dep in pkginfo_target_deps.bzl_select_list(
+                pkg_ctx,
+                target_dep,
+            ):
+                for label in binary_target_dep.value:
+                    if not sets.contains(target_label_set, label):
+                        sets.insert(
+                            binary_target_dep_pairs,
+                            (binary_target_dep.condition, label),
+                        )
 
-    if len(target_labels) == 0:
-        fail("No targets specified for a library product. name:", product.name)
-    deps = [
-        bazel_labels.normalize(label)
-        for label in target_labels
+    binary_target_dep_pairs_list = sets.to_list(binary_target_dep_pairs)
+    unconditional_binary_target_labels = sets.make([
+        label
+        for (condition, label) in binary_target_dep_pairs_list
+        if condition == None
+    ])
+    binary_target_deps = [
+        bzl_selects.new(
+            value = label,
+            kind = pkginfo_target_deps.target_dep_kind,
+            condition = condition,
+        )
+        for (condition, label) in binary_target_dep_pairs_list
+        if condition == None or not sets.contains(
+            unconditional_binary_target_labels,
+            label,
+        )
     ]
     deps.extend(binary_target_deps)
     return build_files.new(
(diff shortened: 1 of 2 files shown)


cgrindel:
Is this related to [the Slack thread](https://bazelbuild.slack.com/archives/C05NZSHE7FU/p1787227155189839) that you started? If so, have you had a chance to look at https://github.com/cgrindel/rules_swift_package_manager/discussions/2488?

cgrindel:
> Is this related to [the Slack thread](https://bazelbuild.slack.com/archives/C05NZSHE7FU/p1787227155189839) that you started? If so, have you had a chance to look at #2488?

Looking at this closer, I see that this is not directly related. Sorry for the noise.

swiftpkg/internal/swiftpkg_build_files.bzl
```diff
@@ -1319,20 +1319,44 @@ def _library_product_build_file(pkg_ctx, product):
...
+                pkginfo_target_deps.bzl_select_list(pkg_ctx, target_dep),
+            )
+
     if len(target_labels) == 0:
         fail("No targets specified for a library product. name:", product.name)
     deps = [
         bazel_labels.normalize(label)
         for label in target_labels
     ]
+    deps.extend(binary_target_deps)
```
cgrindel, line 1379: This can put the same label in `deps` twice, and Bazel treats that as an error.

Two ways it happens:

- Two targets in the same product depend on the same binary target. The loop above appends the label once per target.
- The product lists both a library target and the binary target it depends on (`targets: ["Lib", "MyBinary"]`). `target_labels` already has the binary target, and the loop adds it a second time.

Adding a second member target to the new test fixture produces this:

```
    deps = [
        "@swiftpkg_mypackage//:LibraryWithBinaryTargets.rspm",
        "@swiftpkg_mypackage//:SecondLibWithBinary.rspm",
        "@swiftpkg_mypackage//:BinaryFrameworkTarget.rspm",
        "@swiftpkg_mypackage//:BinaryFrameworkTarget.rspm",
    ] + select({...})
```

Bazel rejects duplicates in both shapes:

- Twice in a plain list: `ERROR: BUILD.bazel:4:8: Label '//:a' is duplicated in the 'deps' attribute of rule 'b'`. The package fails to load, so it fails everywhere.
- Once in the list and once in a `select()` branch: `ERROR: in deps attribute of my_rule rule //:c: Label '//:a' is duplicated`. This one fails at analysis time, and only in the configuration where that branch is picked.

The second shape is the one to worry about. A binary target that is an unconditional dep of one target and an iOS-only dep of another generates a BUILD file that builds fine on macOS and breaks only on iOS.

Deduping the flat list is not enough on its own. A label also has to be dropped from a `select()` branch when it already appears unconditionally, and repeats within a single branch have to go too. Probably easiest to gather `(condition, label)` pairs into a set before building the `bzl_selects` structs, skipping anything already in `target_labels`.

swiftpkg/internal/swiftpkg_build_files.bzl
```diff
@@ -1319,20 +1319,44 @@ def _library_product_build_file(pkg_ctx, product):
...
+    binary_target_deps = []
+    for target in pkg_ctx.pkg_info.targets:
+        if product.name not in target.product_memberships:
+            continue
+        for target_dep in target.dependencies:
+            dep_target_name = None
+            if target_dep.target:
+                dep_target_name = target_dep.target.target_name
+            elif target_dep.by_name:
+                dep_target_name = target_dep.by_name.name
```
cgrindel: A dependency written as `.product(name: "X", package: "MyPkg")` that points at another product in the same package falls through here, since neither `target` nor `by_name` is set. If `X` wraps a binary target, that binary ends up in `X`'s group but not in this one, so the layering check still fails for anyone who imports it through this product.

Fine to leave for a follow-up. Just flagging that the fix does not cover that shape.

swiftpkg/internal/swiftpkg_build_files.bzl
```diff
@@ -1319,20 +1319,44 @@ def _library_product_build_file(pkg_ctx, product):
...
+            if not dep_target_name:
+                continue
+            dep_target = pkginfo_targets.get(
+                targets = pkg_ctx.pkg_info.targets,
+                name = dep_target_name,
+                fail_if_not_found = False,
+            )
```
cgrindel: `pkginfo_targets.get` scans the whole target list on every call, and it now runs inside a loop over products, then targets, then deps. For a package with a lot of all three (Firebase, for instance) that adds real time to the repository rule. Building a `{name: target}` dict once per package would avoid it.

Not introduced by this PR, though. The macro loop around line 91 does the same thing, so take it or leave it.

swiftpkg/tests/swiftpkg_build_files_tests.bzl
```diff
@@ -2090,6 +2144,28 @@ swift_library_group(
     deps = ["@swiftpkg_mypackage//:RegularSwiftTargetAsLibrary.rspm"],
     visibility = ["//visibility:public"],
 )
+""",
+        ),
+        struct(
+            msg = "library product with binary target dependencies",
+            name = "LibraryWithBinaryTargets",
+            pkg_info = _pkg_info(),
```
cgrindel: Could you add two more cases? Both generate output that Bazel will reject today:

- Two targets in the product that depend on the same binary target.
- A product whose `targets` lists both a library target and the binary target it depends on.

swiftpkg/tests/swiftpkg_build_files_tests.bzl
```diff
@@ -554,6 +594,16 @@ def _pkg_info(
...
+            pkginfos.new_target(
+                name = "ConditionalBinaryFrameworkTarget",
+                type = "binary",
+                c99name = "ConditionalBinaryFrameworkTarget",
+                module_type = "BinaryTarget",
+                path = "ConditionalBinaryFrameworkTarget.xcframework",
+                sources = [],
+                dependencies = [],
+                repo_name = _repo_name,
+            ),
```
cgrindel: This target, and the existing `BinaryFrameworkTarget`, have no `product_memberships`. Real data from `swift package describe` is transitive (see `pkginfos.bzl:839`), so a binary target reachable from a product's targets does carry that product in its memberships.

It does not change the result here, since the loop only reads `dependencies` and this target has none. But setting it would make the fixture match what the code actually sees at runtime.

cgrindel, line 751: Set `product_memberships` on both binary target fixtures:

- `ConditionalBinaryFrameworkTarget` (here): `product_memberships = ["LibraryWithBinaryTargets"]`
- `BinaryFrameworkTarget` (line 548): `product_memberships = ["BinaryFrameworkTarget", "LibraryWithBinaryTargets"]`

`swift package describe` reports memberships transitively (see `pkginfos.bzl:839`), so a binary target that a product's target depends on does belong to that product. Without this, the fixture never exercises the new loop reaching a binary target.

swiftpkg/tests/swiftpkg_build_files_tests.bzl
cgrindel, line 2297: Add these two cases. Both hit the duplicate `deps` problem, so they should fail until that is fixed.

Two products, next to `LibraryWithBinaryTargets` above:

```starlark
pkginfos.new_product(
    name = "SharedBinaryProduct",
    type = pkginfos.new_product_type(
        library = pkginfos.new_library_type(
            library_type_kinds.automatic,
        ),
    ),
    targets = ["SharedBinaryLibA", "SharedBinaryLibB"],
),
pkginfos.new_product(
    name = "LibraryWithBinaryInTargets",
    type = pkginfos.new_product_type(
        library = pkginfos.new_library_type(
            library_type_kinds.automatic,
        ),
    ),
    targets = ["LibraryWithBinaryInTargets", "BinaryFrameworkTarget"],
),
```

Three targets to match, each a copy of `LibraryWithBinaryTargets` with only the by-name dep on `BinaryFrameworkTarget`: `SharedBinaryLibA` and `SharedBinaryLibB` with `product_memberships = ["SharedBinaryProduct"]`, and `LibraryWithBinaryInTargets` with `product_memberships = ["LibraryWithBinaryInTargets"]`.

Then the cases themselves:

```starlark
struct(
    msg = "library product where two targets share a binary target dep",
    name = "SharedBinaryProduct",
    pkg_info = _pkg_info(),
    exp = """\
load("@build_bazel_rules_swift//swift:swift.bzl", "swift_library_group")

swift_library_group(
    name = "SharedBinaryProduct",
    deps = [
        "@swiftpkg_mypackage//:SharedBinaryLibA.rspm",
        "@swiftpkg_mypackage//:SharedBinaryLibB.rspm",
        "@swiftpkg_mypackage//:BinaryFrameworkTarget.rspm",
    ],
)
""",
),
struct(
    msg = "library product listing a binary target in its targets",
    name = "LibraryWithBinaryInTargets",
    pkg_info = _pkg_info(),
    exp = """\
load("@build_bazel_rules_swift//swift:swift.bzl", "swift_library_group")

swift_library_group(
    name = "LibraryWithBinaryInTargets",
    deps = [
        "@swiftpkg_mypackage//:LibraryWithBinaryInTargets.rspm",
        "@swiftpkg_mypackage//:BinaryFrameworkTarget.rspm",
    ],
)
""",
),
```

For reference, running these against the current branch gives `BinaryFrameworkTarget.rspm` twice in each `deps` list.

adincebic:
> > Is this related to [the Slack thread](https://bazelbuild.slack.com/archives/C05NZSHE7FU/p1787227155189839) that you started? If so, have you had a chance to look at #2488?
> 
> 
> 
> Looking at this closer, I see that this is not directly related. Sorry for the noise.

Right, this is not related.

cgrindel (changes requested):
Overall, LGTM. Claude and I added a couple of comments.

swiftpkg/internal/swiftpkg_build_files.bzl
adincebic: Makes sense.

adincebic:
> Overall, LGTM. Claude and I added a couple of comments.

Thanks, I addressed them.

cgrindel (approved):
Thank you!

cgrindel:
@mergifyio queue

PR #2511, opened 2026-09-09
feat: make Swift binary injectable through module extension

adincebic:
Because we invoke `swift package` at loading time we can't have rules_swift_package_manager fully use hermetic Swift toolchain.
Here we are allowing the `swift` binary to be injectable so that consumer can provide their own (most likely from standalone Swift toolchain that rules_swift provides).

to make this possible there is only one optional parameter `swift_executable` added to the `from_package(...)`:

```starlark
swift_deps.from_package(
    swift = "//third_party/swift:Package.swift",
    resolved = "//third_party/swift:Package.resolved",
    swift_executable = "@swift_toolchain_ubuntu22.04//:usr/bin/swift",
)
```

Omiting the `swift_executable` parameter will make rules_swift_package_manager fall back to the standard behavior of looking for `swift` binary on the path.

**NOTE:** Standalone Swift toolchain is available in rules_swift 4.x which is currently released as a release candidate but does not affect change in this repo.


cgrindel:
@adincebic I should be able to take a look at this over the weekend.

cgrindel:
> @adincebic I should be able to take a look at this over the weekend.

Actually, I just noticed this is a draft PR. I will review once it is marked ready for review.

PR #2300, opened 2026-05-22
fix: always set minimum_os_version in _spm_minimum_os_transition

john-flanagan:
I noticed in our build that `SwiftDeriveFiles` was being called twice for most of our targets. The cause seems to be a mismatch in the minimum OS version set between one of our unit tests and the primary app target. Those targets set different `minimum_os_version` which weren't being overridden by the `_spm_minimum_os_transition` resulting in differing configs. The `SwiftCompile` action managed to de-dupe the issue, but `SwiftDeriveFiles` did not.

This PR always sets `minimum_os_version` to resolve the issue.

### Example spawns
Notice duplicate `SwiftDeriveFiles`

image

## Validation

You can see the issue if you update the [`minimum_os_version` of `TCAExampleTests`](https://github.com/cgrindel/rules_swift_package_manager/blob/d5b7a68ce65e6dab7eb64f93c8ac24b3eb72005a/examples/tca_example/Tests/BUILD.bazel#L18) to `17.0` and run the `aquery` below.

### Before
```
$ bazel aquery 'mnemonic("SwiftCompile", deps(//Sources:iosapp) union deps(//Tests:TCAExampleTests))' \
      | grep -A 1 "Target: @swiftpkg_swift_composable_architecture//:ComposableArchitecture.rspm.__impl"

  Target: @swiftpkg_swift_composable_architecture//:ComposableArchitecture.rspm.__impl
  Configuration: ios_sim_arm64-fastbuild-ios-sim_arm64-min16.0-ST-4f52eb140c0d
--
  Target: @swiftpkg_swift_composable_architecture//:ComposableArchitecture.rspm.__impl
  Configuration: ios_sim_arm64-fastbuild-ios-sim_arm64-min16.0-ST-98198a139028
```

### After
```
$ bazel aquery 'mnemonic("SwiftCompile", deps(//Sources:iosapp) union deps(//Tests:TCAExampleTests))' \
      | grep -A 1 "Target: @swiftpkg_swift_composable_architecture//:ComposableArchitecture.rspm.__impl"

  Target: @swiftpkg_swift_composable_architecture//:ComposableArchitecture.rspm.__impl
  Configuration: ios_sim_arm64-fastbuild-ios-sim_arm64-min16.0-ST-4f52eb140c0d
```

PR #2512, opened 2026-09-10
feat: propagate package minimum OS floors transitively

### LINKED ISSUE #2509: Likely incorrect propagation of ios_minimum_os parameter trough chain in SPM dependencies
To be clear, the issue was almost completely resolved by AI, so this is what *it* thinks. I was unable to solve it my myself in given time frame.

So at very least error reporting should be improved.
There is also was no test of actual SPM behavior, so that part is not validated, but this is a change in behavior between 0.40.0 and 1.23.0

Following is the report made by AI (qwen3.8 27b) what found the solution for issue happened in the project:

=============

**rspm 1.x per-package minimum OS check hard-fails on package graphs SwiftPM accepts (regression vs 0.x)**

Upgrading rules_swift_package_manager from 0.40 to 1.23.0 breaks analysis with:

```
Error in fail: The target 'SnapKit.rspm' requires ios 12.0, but is being imported from a target configured for ios 10.0; consider raising the importing target's minimum OS to 12.0 or later.
```

**Repro:** depend on Daltron/NotificationBanner v3.2.1 (`platforms: [.iOS(.v10)]`, depends on SnapKit ≥5.0.1 and MarqueeLabel ≥4.0.1, both declaring iOS 12). Any package whose declared minimum is lower than one of its dependencies' declared minimum triggers this — e.g. GoogleDataTransport (iOS 10 → nanopb 12), Firebase iOS SDK 10.x (iOS 11 → nanopb 12).

**What's happening:** rspm 1.x added per-package deployment-target modeling that didn't exist in 0.x. The chain:

1. **BUILD generation** (`swiftpkg/internal/swiftpkg_build_files.bzl`): each SPM target is wrapped in an `spm_minimum_os_target` rule, with attributes like `ios_minimum_os = "10.0"` copied from the package's own `Package.swift` `platforms:` declaration.

2. **The transition** (`swiftpkg/internal/minimum_os_transition.bzl`, `_transition_outputs`): the wrapper's `deps` attribute has `cfg = _spm_minimum_os_transition`. When Bazel configures a dep through it, the transition:
   - takes the package's *own* declared minimum and writes it into the output config (`outputs[option] = minimum_os`) — so each SPM package compiles at its manifest-declared deployment target;
   - **and (the new part)** calls `_import_error(...)`, which `fail()`s if the *importer's* configured minimum is lower than this package's declared minimum.

3. **The failure:** NotificationBannerSwift v3.2.1 declares `.iOS(.v10)` but depends on SnapKit and MarqueeLabel, both declaring iOS 12. So its wrapper configures at 10.0, then importing `SnapKit.rspm` (min 12) trips the check: importer 10.0 < imported 12.0 → hard error at analysis time.

**Why this is a problem:**

- **SwiftPM itself does not enforce this.** A package may declare any platform minimum regardless of its dependencies; mixed deployment targets across modules are legal Swift (each module compiles with its own `-target`; the final app binary builds at the app's higher target). These exact manifests build fine under Xcode/SPM and under rspm 0.40 — so upgrading rspm breaks previously-working graphs.
- **The suggested fix is not actionable.** "Consider raising the importing target's minimum OS" — but the importer here is *another third-party package*, not user code. The only workaround today is patching upstream `Package.swift` files via `swift_deps.configure_package`, which we had to do for three packages (NotificationBanner, GoogleDataTransport, Firebase) in a single app build.
- **The error doesn't name the importer.** "imported from a target configured for ios 10.0" gives no way to find *which* package is misconfigured without reading generated BUILD files; in our case it was `NotificationBannerSwift.rspm.__impl`.

**What would help:** downgrade the import check to a warning (or gate it behind an opt-in flag), or model each package's effective minimum as `max(declared, dependencies' minimums)` and propagate that upward instead of failing — and in any case include the importing target's label in the error message.

=============

Also what is not mentioned there, actual minimal iOS version of the application is iOS 14, set in ios_application rule, so those versions are really just a noise having no relevance for the app itself.

jpsim:
SwiftPM compiles each module against the highest deployment target found among its package and the packages it depends on. The minimum OS transition introduced in #2252/#2259 derived each package's floor from its own `platforms` only, so a target importing a dependency with a higher floor failed the `_import_error` check. Swift 6.4 selects `swift-tools-version: 6.4` manifests, and swift-case-paths 1.10.0 raises its floor to iOS 15 / macOS 12 there while swift-parsing 0.15.2 still declares iOS 13 / macOS 10.15, which broke ordinary dependency bumps:

    Error in fail: The target 'CasePaths.rspm' requires ios 15.0, but is being imported from a target configured for ios 13.0; consider raising the importing target's minimum OS to 15.0 or later.

Fixes #2509, which hits the same check with NotificationBanner 3.2.1 (iOS 10) importing SnapKit and MarqueeLabel (iOS 12), and with GoogleDataTransport and Firebase 10.x importing nanopb, and had to be worked around by patching upstream manifests.

The same move to iOS 15 is spreading through the pointfree graph (swift-dependencies, swift-navigation, swift-perception, swift-sharing, swift-url-routing), so this recurs on routine bumps.

## Approach

Derive the effective floor per target as the maximum of the package's declared (or fallback) version, the floors of the products it imports from other packages, and the floors of the in-package targets it depends on. Deriving per target rather than per package matters: XCMetrics' client target must not inherit macOS 13 from a backend sibling that imports GoogleCloudKit. Each generated repository writes its effective floors per product (and the package-wide maximum) to `minimum_os_versions.json`; a dependent repository reads that file from its sibling repositories (same canonical prefix) so the derivation is transitive without any global pass. Floors are only ever raised.

Referencing an undeclared repository fails a fetch, and SwiftPM's target-based resolution can legitimately leave a dependency out of `Package.resolved`. The `swift_deps` extension therefore declares a `swift_deps_package_repos` hub listing every package repository and passes it to each repository through a new `package_repos` attribute; only listed repositories are read. When the attribute is unset the declared-only behavior is preserved.

`_import_error` stays in place as a safety net, and consumers outside the generated packages configured below a raised floor are still rejected. `minimum_os_version` is still set alongside the platform-specific option (#2300).

#2509 also asks that the error name the importing target; the transition only sees the importer's configuration, not its label, so that is left for a follow-up.

## Validation

- Unit tests for the derivation: chain, diamond, fallback when the dependency omits a platform, dependency declaring a platform the importer does not, never lowering, per-target raising with unrelated targets untouched, unknown-product fallback.
- Analysis test under `swiftpkg/tests/minimum_os_transition/`: an importer wrapper raised to iOS 15 under an iOS 18 consumer builds its `.__impl` and the imported package at 15.
- New `examples/swift_parsing_example`: depends on swift-parsing 0.15.2 (and so swift-case-paths 1.10.0) for the real-world case, plus a local `LocalLow` -> `LocalHigh` package pair with explicit floors so the propagation is exercised on toolchains that select swift-case-paths' Swift 6.0 manifest (GitHub's macOS runners). `do_test` asserts the generated `LocalLow.rspm` wrapper carries `macos_minimum_os = "13.0"` / `ios_minimum_os = "16.0"`. Its `Package.resolved` carries pins for both manifest selections. Verified with Xcode 26.6 and Xcode 27.0 (Swift 6.4); on an unmodified `main` the same example fails with the error above.
- Existing examples re-run: firebase, google_maps, snapkit, messagekit, xcmetrics, resources, sqlite_data, skip_local_transitive_dependencies, module_aliases, pkg_manifest_minimal, and the bzlmod e2e workspace all pass. `bazel test //...` passes.


swiftpkg/internal/minimum_os_versions.bzl
```diff
@@ -25,10 +51,146 @@ def _by_platform(pkg_info):
...
+    versions_by_target = {name: declared for name in target_names}
+
+    def _product_targets_versions(product):
+        return [
+            versions_by_target[tname]
+            for tname in product.targets
+            if tname in versions_by_target
+        ]
+
+    def _dep_versions_list(target_dep):
```
cgrindel: `_dep_versions_list` branches on `target_dep.target` / `.product` / `.by_name` and reads only the name off each ref, never the `condition` field. That field sits on those same refs (`pkginfos.bzl` `_new_product_reference`, `_new_by_name_reference`, `_new_target_reference`), and `pkginfo_target_deps.bzl` reads it to emit the dep inside a `select()`. So a conditional dependency contributes its floors even on platforms where it is not part of the build.

Because `_by_platform` merges per platform key, the effect is cross-platform. Concretely: a package declares iOS 12 and has a dependency it only uses on macOS, and that dependency declares macOS 12 / iOS 14. The iOS 14 floor gets folded into the package's iOS floor, and an app at iOS 13 that builds today now fails with "requires iOS 14.0, but is being imported from a target configured for iOS 13.0; consider raising the importing target's minimum OS." Raising the app's deployment target is the wrong remedy, because the dependency forcing it never enters an iOS build. SwiftPM only considers dependencies that are part of the build for the target platform.

Skipping a dep whose condition excludes all Apple platforms would not be enough, since that still lets a macOS-only dep raise the iOS floor. The fix is to derive per-platform, so a condition contributes only to the platforms it applies to.

One detail for whatever shape that takes: the condition struct also carries `traits`, and per its docstring trait filtering is already applied upstream, with an empty `platforms` list meaning unconditional. So the test wants to key on `platforms`, not on `condition != None`.

swiftpkg/internal/swiftpkg_build_files.bzl
```diff
@@ -1476,12 +1483,18 @@ def _target_visibility(expose_build_targets):
...
         "deps": [actual],
         "visibility": visibility,
     }
-    attrs.update(minimum_os_versions.transition_attrs(pkg_info))
+
+    # Test targets are not derived individually (see
+    # `minimum_os_versions.for_targets`); their wrappers get the package-wide
+    # maximum, which is never lower than any target they depend on.
+    effective = pkg_ctx.minimum_os_versions
+    versions = effective.targets.get(target.name, effective.package)
```
cgrindel: Test targets get the wrong floor in both directions.

Too high: `_for_targets` filters test targets out, so every test wrapper takes this `.get` default — the package-wide maximum. `_transition_outputs` compares the incoming configuration against the wrapper's attr and fails when incoming is lower, so that maximum becomes a hard minimum on how the test may be configured. A package with a Client target at iOS 12 and a Backend target that pulls in something needing iOS 15 can no longer build Client's tests at iOS 12: it fails telling you to raise to iOS 15, because of Backend, which the tests never touch. That is the same complaint this PR fixes for non-test targets.

Too low: `_referenced_dependency_identities` in `pkg_ctxs.bzl` also skips test targets, so a test-only external product's floors are never read at all. The test wrapper gets a package max that does not account for it, transitions to that, and the dependency's own wrapper then fails on the way in — the exact error this PR set out to eliminate, still firing for test targets. That case is also a counterexample to the comment just above: the package-wide maximum is not "never lower than any target they depend on."

Deriving test targets in `_for_targets` fixes the over-raise cleanly, since their unresolved external deps fall back to `[]` and therefore to the declared floors. It only fixes the under-raise if `_referenced_dependency_identities` stops skipping test targets as well, which pulls in more dependency repos at fetch time — worth deciding deliberately rather than as a side effect.

swiftpkg/internal/minimum_os_versions.bzl
cgrindel, line 121: `_dep_versions_list` never looks at a dependency's `condition`, so a dependency used on only one platform raises the floor on all of them.

Say a package declares iOS 12 and has a dependency it only uses on macOS, and that dependency declares iOS 14. The iOS 14 lands on the package's iOS floor, and an app at iOS 13 that builds today fails with "requires iOS 14.0 ... consider raising the importing target's minimum OS" — wrong advice, since that dependency never enters an iOS build.

Skipping deps that exclude Apple entirely wouldn't cover this, since a macOS-only dep still raises iOS. Floors need to be derived per platform. One detail: the condition also carries `traits`, and an empty `platforms` list means unconditional, so key on `platforms`.

swiftpkg/internal/swiftpkg_build_files.bzl
cgrindel, line 1496: Test targets get the wrong floor in both directions.

Too high: test targets are filtered out of `_for_targets`, so every test wrapper falls back to the package-wide max here, and the transition fails any build configured below it. A package with a Client target at iOS 12 and a Backend target needing iOS 15 can't build Client's tests at iOS 12 — it demands iOS 15 because of Backend, which the tests never touch. That's the complaint this PR fixes everywhere else.

Too low: `_referenced_dependency_identities` also skips test targets, so a test-only dependency's floors are never read. The test gets too low a floor and fails deeper in, with the error this PR set out to remove. That case also contradicts the comment above — the package max isn't "never lower than any target they depend on."

Deriving test targets in `_for_targets` fixes the first. The second also needs `_referenced_dependency_identities` to stop skipping them, which costs more fetches.

cgrindel (commented):
Overall, LGTM. Claude flagged a couple of things that may be worth a look before we submit. Let me know what you think.

jpsim:
Thanks for the review @cgrindel! The conditional deps flag was right, fixed that in 2a7f54c6a9db056e181d2587fd4a05e03615441c. The test-target issue isn't actually reachable since `repo_rules` never generates test targets, but the comment there was slightly wrong so I updated that.

cgrindel (approved):
Thank you!

cgrindel:
@mergifyio queue

PR #2519, opened 2026-09-19
fix: raise declared minimum OS to the SwiftPM fallback

cgrindel:
- Raised a declared platform version to the SwiftPM fallback when it is lower, so a package that declares iOS 10 is built and checked at iOS 12. This mirrors `PlatformVersionProvider.getDerived` in SwiftPM, which applies the same floor before `swift build` compares deployment targets.
- Added a unit test covering iOS 10.0 and macOS 10.9. The macOS case guards against a string comparison, which would keep 10.9 over 10.13.

Claude ran `bazel test //swiftpkg/tests/... //swiftpkg/internal/...` and all 313 tests pass. The NotificationBanner repro from the issue was not run.

Related to #2509.


cgrindel:
Closing in favor of #2512, which covers this case and more.

